PR #27, opened 2024-06-05
cnf ran: update pod security label for test namespace

klaskosk:
In the powermanagement test suite, there is a failure due to the stressng pods not meeting the restricted security standard. This updates the namespace to only enforce the baseline standard, which should allow the stressng pods to be created.


klaskosk:
cc: @dgonyier
Leaving this here until we can verify this fixes the issue.

klaskosk:
Closing since this hasn't shown up in multiple weeks of CI. Will reopen if it does come up again.

dgonyier (approved):

PR #28, opened 2024-06-05
cnf ran: talm precache excluded image fix

klaskosk:
To support different image names as they change across versions, the excluded image is now pulled from the Prometheus pod on the cluster. It also only deletes the excluded image if it is found to avoid deleting all spoke images when it is not found.

cc: @yliu127 @josclark42


josclark42:
Sanity check on 4.16 passed. Tests all look good from a functional perspective. 

/lgtm

yliu127 (approved):

trewest (approved):
/lgtm

PR #31, opened 2024-06-07
Draft: Add stability test for ran du

rlmanrique:
Add a stability test for RAN DU.

The idea is to read a set of metrics during some time and store their values in several files. At the end of the test, check those files to see changes.


rlmanrique:
Not ready for review

PR #35, opened 2024-06-10
Added tests for reserved core frequency tuning

josclark42:
Adding a test case for Reserved Core Frequency Tuning.

### DIFF
--- a/tests/cnf/ran/powermanagement/internal/helper/helper.go
+++ b/tests/cnf/ran/powermanagement/internal/helper/helper.go
@@ -3,7 +3,6 @@ package helper
 import (
 	"errors"
 	"fmt"
-	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/cluster"
 	"math"
 	"strconv"
 	"strings"
@@ -14,6 +13,7 @@ import (
 	"github.com/openshift-kni/eco-goinfra/pkg/nodes"
 	"github.com/openshift-kni/eco-goinfra/pkg/nto" //nolint:misspell
 	"github.com/openshift-kni/eco-goinfra/pkg/pod"
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/cluster"
 	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/raninittools"
 	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/powermanagement/internal/tsparams"
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
@@ -79,8 +79,8 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
 	return err
 }
 
-// SetCPUFreqAndWaitForMcpUpdate updates the performance profile with the given isolated and reserved
-// core frequencies and waits for the mcp update.
+// SetCPUFreq updates the performance profile with the given isolated and reserved
+// core frequencies and verifies that the frequencies have been updated on the spoke cluster.
 func SetCPUFreq(
 	perfProfile *nto.Builder, node nodes.Builder,
 	desiredIsolatedCPUFreq *performancev2.CPUfrequency,
@@ -91,32 +91,40 @@ func SetCPUFreq(
 	perfProfile.Definition.Spec.HardwareTuning.IsolatedCpuFreq = desiredIsolatedCPUFreq
 	perfProfile.Definition.Spec.HardwareTuning.ReservedCpuFreq = desiredReservedCPUFreq
 	_, err := perfProfile.Update(true)
+
 	if err != nil {
 		return err
 	}
 
 	// Wait for CPU settings to be applied (work in progress)
-	//time.Sleep(5 * time.Second)
+	// time.Sleep(5 * time.Second)
 
 	// Determine isolated CPU number from IsolatedCPUset
 	isolatedCPUSet, err := cpuset.Parse(string(*perfProfile.Object.Spec.CPU.Isolated))
 	if err != nil {
 		return err
 	}
+
 	isolatedCPUsList := isolatedCPUSet.List()
 	isolatedCPUNumber := isolatedCPUsList[0]
 
-	//Get current isolated core frequency from spoke cluster and compare to desired frequency
+	// Get current isolated core frequency from spoke cluster and compare to desired frequency
 	spokeCommand := fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
 		isolatedCPUNumber)
 	consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+
 	if err != nil {
 		return errors.New(fmt.Sprintf("Command failed: %v", consoleOut))
 	}
+
 	currIsolatedCoreFreq, err := strconv.Atoi(strings.TrimSpace(consoleOut))
 
+	if err != nil {
+		return errors.New(fmt.Sprintf("string conversion failed failed: %v", err))
+	}
+
 	if currIsolatedCoreFreq != int(*desiredIsolatedCPUFreq) {
-		return errors.New("Current Isolated CPU Frequency does not match expected frequency")
+		return errors.New("current Isolated CPU Frequency does not match expected frequency")
 	}
 
 	// Determine reserved CPU number from reservedCPUset
@@ -124,20 +132,26 @@ func SetCPUFreq(
 	if err != nil {
 		return err
 	}
+
 	reservedCPUsList := reservedCPUSet.List()
 	reservedCCPUNumber := reservedCPUsList[0]
 
-	//Get current isolated core frequency from spoke cluster and compare to desired frequency
+	// Get current isolated core frequency from spoke cluster and compare to desired frequency
 	spokeCommand = fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
 		reservedCCPUNumber)
 	consoleOut, err = cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+
 	if err != nil {
-		return errors.New(fmt.Sprintf("Command failed: %v", consoleOut))
+		return errors.New(fmt.Sprintf("command failed: %v", consoleOut))
 	}
 
 	currReservedFreq, err := strconv.Atoi(strings.TrimSpace(consoleOut))
+	if err != nil {
+		return errors.New(fmt.Sprintf("string conversion failed failed: %v", err))
+	}
+
 	if currReservedFreq != int(*desiredReservedCPUFreq) {
-		return errors.New("Current Reserved CPU Frequency does not match expected frequency")
+		return errors.New("current Reserved CPU Frequency does not match expected frequency")
 	}
 
 	return err
--- a/tests/cnf/ran/powermanagement/internal/helper/helper.go
+++ b/tests/cnf/ran/powermanagement/internal/helper/helper.go
@@ -147,14 +147,8 @@ func SetCPUFreq(
 				return false, nil
 			}
 
-			return true, nil
-		})
-
-	// Wait for Reserved CPU Frequency to be updated.
-	err = wait.PollUntilContextTimeout(context.TODO(), 5*time.Second, 1*time.Minute, true,
-		func(ctx context.Context) (bool, error) {
 			// Get current isolated core frequency from spoke cluster and compare to desired frequency
-			cmdOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommandReservedCPUs)
+			cmdOut, err = cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommandReservedCPUs)
 
 			if err != nil {
 				return false, fmt.Errorf("command failed: %s", cmdOut)
--- a/tests/cnf/ran/powermanagement/internal/helper/helper.go
+++ b/tests/cnf/ran/powermanagement/internal/helper/helper.go
@@ -120,13 +120,13 @@ func SetCPUFreq(
 			consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
 
 			if err != nil {
-				return false, errors.New(fmt.Sprintf("Command failed: %v", consoleOut))
+				return false, fmt.Errorf("Command failed: %w", consoleOut)
 			}
 
 			currIsolatedCoreFreq, err := strconv.Atoi(strings.TrimSpace(consoleOut))
 
 			if err != nil {
-				return false, errors.New(fmt.Sprintf("string conversion failed failed: %v", err))
+				return false, fmt.Errorf("string conversion failed failed: %v", err)
 			}
 
 			if currIsolatedCoreFreq != int(*desiredIsolatedCoreFreq) {
@@ -155,12 +155,12 @@ func SetCPUFreq(
 			consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
 
 			if err != nil {
-				return false, errors.New(fmt.Sprintf("command failed: %v", consoleOut))
+				return false, fmt.Errorf("command failed: %v", consoleOut)
 			}
 
 			currReservedFreq, err := strconv.Atoi(strings.TrimSpace(consoleOut))
 			if err != nil {
-				return false, errors.New(fmt.Sprintf("string conversion failed failed: %v", err))
+				return false, fmt.Errorf("string conversion failed failed: %v", err)
 			}
 
 			if currReservedFreq != int(*desiredReservedCoreFreq) {
--- a/tests/cnf/ran/powermanagement/internal/helper/helper.go
+++ b/tests/cnf/ran/powermanagement/internal/helper/helper.go
@@ -82,7 +82,7 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
 // SetCPUFreq updates the performance profile with the given isolated and reserved
 // core frequencies and verifies that the frequencies have been updated on the spoke cluster.
 func SetCPUFreq(
-	perfProfile *nto.Builder, node nodes.Builder,
+	perfProfile *nto.Builder,
 	desiredIsolatedCPUFreq *performancev2.CPUfrequency,
 	desiredReservedCPUFreq *performancev2.CPUfrequency) error {
 	glog.V(tsparams.LogLevel).Infof("Set Reserved and Isolated CPU Frequency on performance profile")
@@ -97,7 +97,7 @@ func SetCPUFreq(
 	}
 
 	// Wait for CPU settings to be applied (work in progress)
-	// time.Sleep(5 * time.Second)
+	time.Sleep(5 * time.Second)
 
 	// Determine isolated CPU number from IsolatedCPUset
 	isolatedCPUSet, err := cpuset.Parse(string(*perfProfile.Object.Spec.CPU.Isolated))
--- a/tests/cnf/ran/powermanagement/internal/helper/helper.go
+++ b/tests/cnf/ran/powermanagement/internal/helper/helper.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"math"
 	"strconv"
 	"strings"
@@ -89,6 +90,8 @@ func SetCPUFreq(
 	desiredReservedCoreFreq *performancev2.CPUfrequency) error {
 	glog.V(tsparams.LogLevel).Infof("Set Reserved and Isolated CPU Frequency on performance profile")
 
+	//perfProfile, err := GetPerformanceProfileWithCPUSet()
+
 	// Update PerfProfile with new CPU Frequencies
 	perfProfile.Definition.Spec.HardwareTuning.IsolatedCpuFreq = desiredIsolatedCoreFreq
 	perfProfile.Definition.Spec.HardwareTuning.ReservedCpuFreq = desiredReservedCoreFreq
@@ -104,11 +107,22 @@ func SetCPUFreq(
 		return err
 	}
 
+	// Determine reserved CPU number from reservedCPUset
+	reservedCPUSet, err := cpuset.Parse(string(*perfProfile.Object.Spec.CPU.Reserved))
+	if err != nil {
+		return err
+	}
+
 	isolatedCPUsList := isolatedCPUSet.List()
 	isolatedCPUNumber := isolatedCPUsList[0]
+	reservedCPUsList := reservedCPUSet.List()
+	reservedCCPUNumber := reservedCPUsList[0]
 
-	spokeCommand := fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
+	spokeCommandIsolatedCPUs := fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq",
 		isolatedCPUNumber)
+
+	spokeCommandReservedCPUs := fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq",
+		reservedCCPUNumber)
 
 	// Wait for Isolated CPU Frequency to be updated.
 	err = wait.PollUntilContextTimeout(
@@ -117,7 +131,7 @@ func SetCPUFreq(
 				return false, err
 			}
 			// Get current isolated core frequency from spoke cluster and compare to desired frequency
-			cmdOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+			cmdOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommandIsolatedCPUs)
 
 			if err != nil {
 				return false, fmt.Errorf("command failed: %s", cmdOut)
@@ -136,23 +150,11 @@ func SetCPUFreq(
 			return true, nil
 		})
 
-	// Determine reserved CPU number from reservedCPUset
-	reservedCPUSet, err := cpuset.Parse(string(*perfProfile.Object.Spec.CPU.Reserved))
-	if err != nil {
-		return err
-	}
-
-	reservedCPUsList := reservedCPUSet.List()
-	reservedCCPUNumber := reservedCPUsList[0]
-
-	spokeCommand = fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
-		reservedCCPUNumber)
-
 	// Wait for Reserved CPU Frequency to be updated.
 	err = wait.PollUntilContextTimeout(context.TODO(), 5*time.Second, 1*time.Minute, true,
 		func(ctx context.Context) (bool, error) {
 			// Get current isolated core frequency from spoke cluster and compare to desired frequency
-			cmdOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+			cmdOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommandReservedCPUs)
 
 			if err != nil {
 				return false, fmt.Errorf("command failed: %s", cmdOut)
@@ -166,6 +168,8 @@ func SetCPUFreq(
 			if currReservedFreq != int(*desiredReservedCoreFreq) {
 				return false, nil
 			}
+
+			log.Println("****NEW RCF: %v   ******", currReservedFreq)
 
 			return true, nil
 		})
--- a/tests/cnf/ran/powermanagement/internal/helper/helper.go
+++ b/tests/cnf/ran/powermanagement/internal/helper/helper.go
@@ -3,8 +3,10 @@ package helper
 import (
 	"errors"
 	"fmt"
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/cluster"
 	"math"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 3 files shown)


josclark42:
@yliu127 This is the official PR for Reserved Core Frequency Tuning. I already received feedback from @klaskosk. Please let me know if either of you have suggestions or questions.

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +77,28 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+
+	perfProfile.Definition.Spec.HardwareTuning.IsolatedCpuFreq = isolatedCPUFreq
+	perfProfile.Definition.Spec.HardwareTuning.ReservedCpuFreq = reservedCPUFreq
+
+	_, err := perfProfile.Update(true)
+	if err != nil {
+		return err
+	}
+
+	// Wait for CPU settings to be applied (work in progress)
```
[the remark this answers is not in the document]
yliu127: Does this change trigger server reboot?

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+var _ = Describe("Per-core runtime power states tuning", Label(tsparams.LabelCPUFrequency), func() {
+	var (
+		nodeList                []*nodes.Builder
+		perfProfile             *nto.Builder
+		err                     error
+		desiredReservedCoreFreq = performancev2.CPUfrequency(2500002)
+		desiredIsolatedCoreFreq = performancev2.CPUfrequency(2200002)
+		originalIsolatedCPUFreq performancev2.CPUfrequency
+		originalReservedCPUFreq performancev2.CPUfrequency
+		isolatedCPUNumber       = 2
```
yliu127, line 28: this shouldn't be hardcoded. should take config from the same perf profile to get isolated cpu list and pick one cpu from the list.

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+		ReservedCPUNumber       = 0
```
yliu127, line 29: this shouldn't be hardcoded.

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+
+			By("Get modified isolated core frequency")
+			spokeCommand := fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
+				isolatedCPUNumber)
+			consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+			Expect(err).ToNot(HaveOccurred(), "Failed to %s", spokeCommand)
+
+			By("Compare current isolated core freq to desired isolated core freq")
+			currIsolatedCoreFreq, err := strconv.Atoi(strings.Trim(consoleOut, "\r\n "))
+			Expect(err).ToNot(HaveOccurred(), "strconv.Atoi Failed")
```
yliu127, line 82: consoleOut should be included in error string to help debugging in case failure happens here. As it indicates the cmd output cannot be converted to a number, so it's important to know what is the exact output,

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+	FContext("Reserved Core Frequency Tuning Test", func() {
+		It("tests changing reserved and isolated CPU frequencies", func() {
+
+			By("patch performance profile to set core frequencies")
+			err = helper.SetCPUFreqAndWaitForMcpUpdate(perfProfile, *nodeList[0],
+				&desiredIsolatedCoreFreq, &desiredReservedCoreFreq)
+			Expect(err).ToNot(HaveOccurred(), "Failed to set CPU Freq")
+
+			By("Get modified isolated core frequency")
+			spokeCommand := fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
```
yliu127, line 75: why adding "| cat -" to mask cmd failure? It would be better to let the test fail when the actual issue is encountered. 

```
[kni@registry.kni-qe-72 ~]$ cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -
cat: /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq: No such file or directory
[kni@registry.kni-qe-72 ~]$ echo $?
0
[kni@registry.kni-qe-72 ~]$ cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq
cat: /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq: No such file or directory
[kni@registry.kni-qe-72 ~]$ echo $?
1
```

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+	})
+
+	FContext("Reserved Core Frequency Tuning Test", func() {
+		It("tests changing reserved and isolated CPU frequencies", func() {
+
+			By("patch performance profile to set core frequencies")
+			err = helper.SetCPUFreqAndWaitForMcpUpdate(perfProfile, *nodeList[0],
+				&desiredIsolatedCoreFreq, &desiredReservedCoreFreq)
+			Expect(err).ToNot(HaveOccurred(), "Failed to set CPU Freq")
+
```
yliu127, line 73: It may be easier to include the verification steps into the SetCPUFreq helper. Because the scope of the helper, it makes sense that way. It also helps verify the revert succeeded end to end in AfterXX.

yliu127 (changes requested):

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/cluster"
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/raninittools"
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/powermanagement/internal/helper"
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/powermanagement/internal/tsparams"
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+)
+
+var _ = Describe("Per-core runtime power states tuning", Label(tsparams.LabelCPUFrequency), func() {
+	var (
+		nodeList                []*nodes.Builder
```
yliu127, line 21: should remove. This is only used in BeforeXX.

These vars are useful if you set and retrieve the value in different functions. e.g., Set those values once in BeforeAll, and use them in the test function.

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+		perfProfile             *nto.Builder
+		err                     error
```
yliu127, line 23: is this needed?

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/openshift-kni/eco-goinfra/pkg/nodes"
+	"github.com/openshift-kni/eco-goinfra/pkg/nto" //nolint:misspell
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/cluster"
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/raninittools"
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/powermanagement/internal/helper"
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/powermanagement/internal/tsparams"
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+)
+
+var _ = Describe("Per-core runtime power states tuning", Label(tsparams.LabelCPUFrequency), func() {
```
yliu127, line 19: This is exactly the same as existing test suite powersave.go, which indicate we should not create separate file for this new test.

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+		desiredReservedCoreFreq = performancev2.CPUfrequency(2500002)
+		desiredIsolatedCoreFreq = performancev2.CPUfrequency(2200002)
+		originalIsolatedCPUFreq performancev2.CPUfrequency
+		originalReservedCPUFreq performancev2.CPUfrequency
+		isolatedCPUNumber       = 2
+		ReservedCPUNumber       = 0
+	)
+
+	BeforeEach(func() {
+		nodeList, err = nodes.List(raninittools.Spoke1APIClient)
```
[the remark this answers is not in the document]
klaskosk: We don't, I wrote the tests assuming they just wouldn't be run for non SNO clusters. Probably need to update some to not fail but instead skip if the cluster is not SNO then

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+		err                     error
+		desiredReservedCoreFreq = performancev2.CPUfrequency(2500002)
+		desiredIsolatedCoreFreq = performancev2.CPUfrequency(2200002)
+		originalIsolatedCPUFreq performancev2.CPUfrequency
+		originalReservedCPUFreq performancev2.CPUfrequency
+		isolatedCPUNumber       = 2
+		ReservedCPUNumber       = 0
+	)
+
+	BeforeEach(func() {
```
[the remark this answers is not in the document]
klaskosk: The BeforeEach instead of BeforeAll was my suggestion since there's only one test in the file the ordered is a bit redundant but if the BeforeEach is expensive then it'd make sense to use continue on failure. My understanding is there's no reboot in this test though

tests/cnf/ran/powermanagement/tests/cpufreq.go
yliu127: This is actually used in AfterXX and It. So closing this comment.

tests/cnf/ran/powermanagement/tests/cpufreq.go
yliu127: @klaskosk It will make retry pretty much useless if we append "| cat - ". How often do you encounter this issue? We might need to think of other solution to strip off the timestamp.

tests/cnf/ran/powermanagement/tests/cpufreq.go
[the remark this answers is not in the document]
yliu127: @klaskosk right, but we do revert the config in AfterXX, so best to do that in AfterAll I think if we will have multiple  test cases in the future. But we can also change this structure when we actually have more tests in the future. @josclark42  feel free to keep Before/AfterEach for this PR if you would prefer that way.

tests/cnf/ran/powermanagement/tests/cpufreq.go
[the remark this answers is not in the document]
yliu127: @klaskosk yes we need to change that assumption. 

Default is all tests can be run on any type of clusters. And if specific tests are for SNO only, then we would skip them when non-SNO clusters are detected. 

@josclark42 let's keep this one as is.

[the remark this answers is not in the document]
klaskosk: @yliu127 Okay I'll have to do a review of some of the existing tests to add that in

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/openshift-kni/eco-goinfra/pkg/nodes"
+	"github.com/openshift-kni/eco-goinfra/pkg/nto" //nolint:misspell
```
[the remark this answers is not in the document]
josclark42: @yliu127 Without this comment lint fails with the errror:
`tests/cnf/ran/powermanagement/tests/cpufreq.go:10:44: `nto` is a misspelling of `not` (misspell)`

josclark42:
@yliu127 @klaskosk I made some major changes which resolved a lot of comments and invalidated others. Notably:

- moved the CPU frequency verification from the test to the helper function.
- Added a polling function to the helper to wait for the change to take effect
- Removed unneeded nodelist variable
- removed nodelist as a parameter for the helper function
-  other small changes as per review recommendations.

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+			consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+
+			if err != nil {
+				return false, errors.New(fmt.Sprintf("Command failed: %v", consoleOut))
+			}
+
+			currIsolatedCoreFreq, err := strconv.Atoi(strings.TrimSpace(consoleOut))
+
+			if err != nil {
+				return false, errors.New(fmt.Sprintf("string conversion failed failed: %v", err))
```
klaskosk, line 129: Can use `fmt.Errorf` with the `%w` verb here instead of `%v`

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+	err = wait.PollUntilContextTimeout(
+		context.TODO(), 5*time.Second, 1*time.Minute, true, func(ctx context.Context) (bool, error) {
+			if err != nil {
+				return false, err
+			}
+			// Get current isolated core frequency from spoke cluster and compare to desired frequency
+			consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+
+			if err != nil {
+				return false, errors.New(fmt.Sprintf("Command failed: %v", consoleOut))
```
klaskosk, line 123: Can use `fmt.Errorf` with the `%w` verb here instead of `%v`

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+		reservedCCPUNumber)
+
+	// Wait for Reserved CPU Frequency to be updated.
+	err = wait.PollUntilContextTimeout(context.TODO(), 5*time.Second, 1*time.Minute, true,
+		func(ctx context.Context) (bool, error) {
+			// Get current isolated core frequency from spoke cluster and compare to desired frequency
+			consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+
+			if err != nil {
+				return false, errors.New(fmt.Sprintf("command failed: %v", consoleOut))
```
klaskosk, line 158: See above comments for all errors

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,83 @@
...
+		originalReservedCPUFreq performancev2.CPUfrequency
+	)
+
+	BeforeEach(func() {
+		var err error
+
+		perfProfile, err = helper.GetPerformanceProfileWithCPUSet()
+		Expect(err).ToNot(HaveOccurred(), "Failed to get performance profile")
+
+		// Get isolated core ID
```
klaskosk, line 34:
```suggestion
		By("getting isolated core ID")
```

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,83 @@
...
+		isolatedCPUSet, err := cpuset.Parse(string(*perfProfile.Object.Spec.CPU.Isolated))
+		Expect(err).ToNot(HaveOccurred(), "Failed to get isolated cpu set")
+		isolatedCPUsList := isolatedCPUSet.List()
+		isolatedCPUNumber := isolatedCPUsList[0]
+
+		// Get reserved core ID
```
klaskosk, line 40:
```suggestion
		By("getting reserved core ID")
```

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,83 @@
...
+		Expect(err).ToNot(HaveOccurred(), "Failed to %s, error: %s", spokeCommand, consoleOut)
+		freqAsInt, err = strconv.Atoi(strings.TrimSpace(consoleOut))
+		Expect(err).ToNot(HaveOccurred(), "strconv.Atoi Failed")
+		originalReservedCPUFreq = performancev2.CPUfrequency(freqAsInt)
+
+	})
+
+	AfterEach(func() {
+		By("Reverts the CPU frequencies to the original setting")
+		err := helper.SetCPUFreq(perfProfile,
```
klaskosk, line 68: can this be one line?

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,83 @@
...
+			&originalIsolatedCPUFreq, &originalReservedCPUFreq)
+		Expect(err).ToNot(HaveOccurred(), "Failed to set CPU Freq")
+	})
+
+	Context("Reserved Core Frequency Tuning Test", func() {
+
+		It("tests changing reserved and isolated CPU frequencies", func() {
+			By("patch performance profile to set core frequencies")
+			err := helper.SetCPUFreq(perfProfile,
```
klaskosk, line 77: Same as above, can this be one line?

tests/cnf/ran/powermanagement/internal/tsparams/consts.go
```diff
@@ -11,7 +11,9 @@ const (
 	LabelSuite = "powermanagement"
 	// LabelPowerSaveTestCases is the label for a particular test case.
 	LabelPowerSaveTestCases = "powersave"
-	// PowerSaveTimeout is the timeout being used in powersave tests.
+	// LabelCPUFrequency is the label for frequency tuning test cases.
+	LabelCPUFrequency = "cpu_frequency"
```
klaskosk, line 15:
```suggestion
	LabelCPUFrequency = "cpu-frequency"
```
I've been using hyphens so this is just for consistency

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+			}
+
+			currIsolatedCoreFreq, err := strconv.Atoi(strings.TrimSpace(consoleOut))
+
+			if err != nil {
+				return false, errors.New(fmt.Sprintf("string conversion failed failed: %v", err))
+			}
+
+			if currIsolatedCoreFreq != int(*desiredIsolatedCoreFreq) {
+				return false, errors.New("current Isolated CPU Frequency does not match expected frequency")
```
klaskosk, line 133: returning false with a non-nil error fails the `wait.PollUntilContextTimeout` so this might need to be `return false, nil`

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+				return false, errors.New(fmt.Sprintf("command failed: %v", consoleOut))
+			}
+
+			currReservedFreq, err := strconv.Atoi(strings.TrimSpace(consoleOut))
+			if err != nil {
+				return false, errors.New(fmt.Sprintf("string conversion failed failed: %v", err))
+			}
+
+			if currReservedFreq != int(*desiredReservedCoreFreq) {
+				return false, errors.New("current Reserved CPU Frequency does not match expected frequency")
```
klaskosk, line 167: Same as above about non-nil error returns

klaskosk (changes requested):
A few mostly style comments

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: Done.

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: Done.

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: Done.

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: Done

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: Done.

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: Good idea- thanks. Done.

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: Done- thanks.

tests/cnf/ran/powermanagement/internal/tsparams/consts.go
josclark42: Thank you done.

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+
+		perfProfile, err = helper.GetPerformanceProfileWithCPUSet()
+		Expect(err).ToNot(HaveOccurred(), "Failed to get performance profile")
+
+		By("get original isolated core frequency")
+		spokeCommand := fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
+			isolatedCPUNumber)
+		consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+		Expect(err).ToNot(HaveOccurred(), "Failed to %s", spokeCommand)
+		freqAsInt, err := strconv.Atoi(strings.Trim(consoleOut, "\r\n"))
```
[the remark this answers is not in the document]
josclark42: all strings.Trim have been replaced with strings.TrimSpace

tests/cnf/ran/powermanagement/internal/helper/helper.go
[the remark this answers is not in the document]
josclark42: Changed to PollUntilContextTimeout

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: I updated the test description.

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: Kirsten pointed out that it is needed, though not in the global variables. If := is used when getting the perfProfile, the perfProfile var is only scoped to within the beforeEach which causes problems when this var is used later.

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: I refactored the helper to do the varification.

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: I updated the error to include the console output.

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: As per your suggestion, I changed this to get the CPUSets from the PerfProfile, and select the first  CPU from each of the two arrays.

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: As per your suggestion, I changed this to get the CPUSets from the PerfProfile, and select the first CPU from each of the two arrays.

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,100 @@
...
+		originalIsolatedCPUFreq performancev2.CPUfrequency
+		originalReservedCPUFreq performancev2.CPUfrequency
+		isolatedCPUNumber       = 2
+		ReservedCPUNumber       = 0
+	)
+
+	BeforeEach(func() {
+		nodeList, err = nodes.List(raninittools.Spoke1APIClient)
+		Expect(err).ToNot(HaveOccurred(), "Failed to get nodes")
+		Expect(len(nodeList)).To(Equal(1), "Currently only SNO clusters are supported")
```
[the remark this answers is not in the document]
josclark42: nodeList is not needed at all for this test so I removed it.

tests/cnf/ran/powermanagement/internal/helper/helper.go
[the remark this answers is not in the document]
josclark42: I changed this to a "PollUntilContextTimeout"

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: @klaskosk  I changed this to a fmt.Errorf. I don't think there's a reason to use errors.New.

tests/cnf/ran/powermanagement/internal/helper/helper.go
klaskosk: @josclark42 In this case it's not about `fmt.Errorf` vs `errors.New`, it's that returning an error from `wait.PollUntilContextTimeout` will prevent it from trying again after the interval. I think in this case you still want to keep checking if the current is equal to the desired so you should return a nil error. Otherwise the function will only run once.

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+		reservedCCPUNumber)
+
+	// Wait for Reserved CPU Frequency to be updated.
+	err = wait.PollUntilContextTimeout(context.TODO(), 5*time.Second, 1*time.Minute, true,
+		func(ctx context.Context) (bool, error) {
+			// Get current isolated core frequency from spoke cluster and compare to desired frequency
+			consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+
+			if err != nil {
+				return false, fmt.Errorf("command failed: %q", consoleOut)
```
klaskosk, line 158: I'm not familiar with `%q`, is there a reason to use it over `%s` here?

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+	err = wait.PollUntilContextTimeout(
+		context.TODO(), 5*time.Second, 1*time.Minute, true, func(ctx context.Context) (bool, error) {
+			if err != nil {
+				return false, err
+			}
+			// Get current isolated core frequency from spoke cluster and compare to desired frequency
+			consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+
+			if err != nil {
+				return false, fmt.Errorf("command failed: %q", consoleOut)
```
klaskosk, line 123: Same question as below

tests/cnf/ran/powermanagement/tests/cpufreq.go
klaskosk: It looks like this is still multiple lines

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: Got it now. Thanks for the explanation! Fixed.

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: Fixed

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: In an earlier version lint was complaining about the %s so I changed to %q which satisfied it. I presume it has something to do with the output containing characters that must be double-quoted? 

Regardless, it now passes with a %s which I've changed to for the sake of consistency.

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: In an earlier version lint was complaining about the %s so I changed to %q which satisfied it. I presume it has something to do with the output containing characters that must be double-quoted? 

Regardless, it now passes with a %s which I've changed to for the sake of consistency.

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: Fixed for real this time?

tests/cnf/ran/powermanagement/tests/cpufreq.go
yliu127: yes thats what I meant as well. It is not needed as a shared/global var.  Thanks for the change.

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+		isolatedCPUNumber)
+
+	// Wait for Isolated CPU Frequency to be updated.
+	err = wait.PollUntilContextTimeout(
+		context.TODO(), 5*time.Second, 1*time.Minute, true, func(ctx context.Context) (bool, error) {
+			if err != nil {
+				return false, err
+			}
+			// Get current isolated core frequency from spoke cluster and compare to desired frequency
+			consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
```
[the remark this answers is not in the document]
josclark42: Done- thank you.

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,81 @@
...
+
+		})
+
+		AfterEach(func() {
+			By("reverting the CPU frequencies to the original setting")
+			err := helper.SetCPUFreq(perfProfile, &originalIsolatedCPUFreq, &originalReservedCPUFreq)
+			Expect(err).ToNot(HaveOccurred(), "Failed to set CPU Freq")
+		})
+
+		Context("Reserved Core Frequency Tuning Test", func() {
```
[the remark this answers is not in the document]
josclark42: Done- thank you.

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+	// Determine isolated CPU number from IsolatedCPUset
+	isolatedCPUSet, err := cpuset.Parse(string(*perfProfile.Object.Spec.CPU.Isolated))
+	if err != nil {
+		return err
+	}
+
+	isolatedCPUsList := isolatedCPUSet.List()
+	isolatedCPUNumber := isolatedCPUsList[0]
+
+	spokeCommand := fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
```
klaskosk, line 110: After a bit of discussion I think the `| cat -` can be removed from this function since the command gets retried so malformed output occasionally isn't an issue

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+	// Determine reserved CPU number from reservedCPUset
+	reservedCPUSet, err := cpuset.Parse(string(*perfProfile.Object.Spec.CPU.Reserved))
+	if err != nil {
+		return err
+	}
+
+	reservedCPUsList := reservedCPUSet.List()
+	reservedCCPUNumber := reservedCPUsList[0]
+
+	spokeCommand = fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
```
klaskosk, line 148: See above about removing `| cat -` in this function specifically

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,81 @@
...
+		})
+
+		AfterEach(func() {
+			By("reverting the CPU frequencies to the original setting")
+			err := helper.SetCPUFreq(perfProfile, &originalIsolatedCPUFreq, &originalReservedCPUFreq)
+			Expect(err).ToNot(HaveOccurred(), "Failed to set CPU Freq")
+		})
+
+		When("reserved and isolated core frequency is configured via PerformanceProfile", func() {
+
```
klaskosk, line 74: nitpick, but this blank line can be removed

klaskosk (commented):
Some more clarity on the `| cat -` situation in the helper

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+
+	isolatedCPUsList := isolatedCPUSet.List()
+	isolatedCPUNumber := isolatedCPUsList[0]
+
+	spokeCommand := fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
+		isolatedCPUNumber)
+
+	// Wait for Isolated CPU Frequency to be updated.
+	err = wait.PollUntilContextTimeout(
+		context.TODO(), 5*time.Second, 1*time.Minute, true, func(ctx context.Context) (bool, error) {
```
klaskosk, line 115: Perhaps consider reducing the interval too if it's a quick update since we may have to wait for a couple intervals if the output is malformed

tests/cnf/ran/powermanagement/internal/helper/helper.go
```diff
@@ -77,6 +81,98 @@ func SetPowerModeAndWaitForMcpUpdate(perfProfile *nto.Builder, node nodes.Builde
...
+	// Determine reserved CPU number from reservedCPUset
+	reservedCPUSet, err := cpuset.Parse(string(*perfProfile.Object.Spec.CPU.Reserved))
+	if err != nil {
+		return err
+	}
+
+	reservedCPUsList := reservedCPUSet.List()
+	reservedCCPUNumber := reservedCPUsList[0]
+
+	spokeCommand = fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
```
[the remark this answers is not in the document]
josclark42: Wrapper created.

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -0,0 +1,81 @@
...
+			spokeCommand := fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
+				isolatedCPUNumber)
+			consoleOut, err := cluster.ExecCommandOnSNO(raninittools.Spoke1APIClient, 3, spokeCommand)
+			Expect(err).ToNot(HaveOccurred(), "Failed to %s, error:%s", spokeCommand, consoleOut)
+			freqAsInt, err := strconv.Atoi(strings.TrimSpace(consoleOut))
+			Expect(err).ToNot(HaveOccurred(), "strconv.Atoi Failed")
+			originalIsolatedCPUFreq = performancev2.CPUfrequency(freqAsInt)
+
+			By("getting original reserved core frequency")
+			spokeCommand = fmt.Sprintf("cat /sys/devices/system/cpu/cpufreq/policy%v/scaling_max_freq |cat -",
```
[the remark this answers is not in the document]
josclark42: Wrapper created.

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: Removed.

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: Removed- thanks.

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: Removed.

tests/cnf/ran/powermanagement/internal/helper/helper.go
josclark42: Reduced interval to 2

yliu127 (approved):
/lgtm

klaskosk (approved):
/lgtm

PR #43, opened 2024-06-12
cnf ran: remove cat hack from powersave

klaskosk:
The powersave test cases would pipe the output from cat into `cat -` which causes errors to be ignored as a workaround for ocassional malformed output. This has been converted to us Eventually to allow for retrying on malformed output.


klaskosk:
I've successfully tested this change on 4.15

klaskosk:
Tested this change on 4.16.0 and verified it worked

yliu127 (approved):

mpmaruthu (approved):

PR #50, opened 2024-06-17
lca: add negative test for patching the stage from prep to rollback.

achuzhoy:


tests/lca/imagebasedupgrade/mgmt/negative/tests/ibu-invalid-next-stage.go
```diff
@@ -68,4 +68,24 @@ var _ = Describe(
 				"error: ibu seedimage updated with wrong next stage")
 
 		})
+
+		FIt("fails because from Prep it's not possible to move to Rollback stage", reportxml.ID("71740"), func() {
```
trewest, line 72:
```suggestion
		It("fails because from Prep it's not possible to move to Rollback stage", reportxml.ID("71740"), func() {
```

tests/assisted/ztp/operator/tests/lca_sasha.go
```diff
@@ -0,0 +1,134 @@
+package operator_test
```
trewest, line 1: Leftover

tests/lca/imagebasedupgrade/mgmt/negative/tests/ibu-invalid-next-stage.go
achuzhoy: Updated. thanks.

tests/assisted/ztp/operator/tests/lca_sasha.go
achuzhoy: removed. thanks

trewest (approved):
/lgtm

PR #51, opened 2024-06-18
Add Version Check to Reserved Core Freqency Tuning Test. Fix and move CompareOCPVersionWithCurrent() 

josclark42:
- This PR adds an OCP version check to the Reserved Core Freqency Tuning test, ensuring it only runs on OCP 4.16 or greater.

- CompareOCPVersionWithCurrent() was moved from  tests/system-tests/internal/platform/platform.go to tests/internal/cluster/config.go so it could be imported into tests/cnf/ran/tests
 
- tests/system-tests/vcore/internal/vcorecommon/cgroup-operator.go was updated to reflect the new location of CompareOCPVersionWithCurrent()
 
- Finally, a bug was found and fixed in CompareOCPVersionWithCurrent(). Previously, in the case where isGreater=true and the check failed, the LessThan function would still run and the function would return true erroneously. Essentially the function was testing for both greater than and less than at the same time.  I added an if !greater block around the final check to fix this.

### DIFF
--- a/tests/cnf/ran/powermanagement/tests/cpufreq.go
+++ b/tests/cnf/ran/powermanagement/tests/cpufreq.go
@@ -9,6 +9,7 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/openshift-kni/eco-goinfra/pkg/nto" //nolint:misspell
 	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/cluster"
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/ranhelper"
 	. "github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/raninittools"
 	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/powermanagement/internal/helper"
 	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/powermanagement/internal/tsparams"
@@ -63,7 +64,15 @@ var _ = Describe("CPU frequency tuning tests change the core frequencies of isol
 		When("reserved and isolated core frequency is configured via PerformanceProfile", func() {
 
 			It("sets the reserved and isolated core frequency correctly on the DUT", func() {
-				err := helper.SetCPUFreq(perfProfile, &desiredIsolatedCoreFreq, &desiredReservedCoreFreq)
+
+				versionInRange, err := ranhelper.IsVersionStringInRange(RANConfig.Spoke1OCPVersion, "4.16", "")
+				Expect(err).ToNot(HaveOccurred(), "Failed to compare OCP version string")
+
+				if !versionInRange {
+					Skip("OCP 4.16 or higher required for this test")
+				}
+
+				err = helper.SetCPUFreq(perfProfile, &desiredIsolatedCoreFreq, &desiredReservedCoreFreq)
 				Expect(err).ToNot(HaveOccurred(), "Failed to set CPU Freq")
 
 			})
--- a/tests/cnf/ran/powermanagement/internal/helper/helper.go
+++ b/tests/cnf/ran/powermanagement/internal/helper/helper.go
@@ -89,6 +89,10 @@ func SetCPUFreq(
 	desiredReservedCoreFreq *performancev2.CPUfrequency) error {
 	glog.V(tsparams.LogLevel).Infof("Set Reserved and Isolated CPU Frequency on performance profile")
 
+	if perfProfile.Definition.Spec.HardwareTuning == nil {
+		perfProfile.Definition.Spec.HardwareTuning = &performancev2.HardwareTuning{}
+	}
+
 	// Update PerfProfile with new CPU Frequencies
 	perfProfile.Definition.Spec.HardwareTuning.IsolatedCpuFreq = desiredIsolatedCoreFreq
 	perfProfile.Definition.Spec.HardwareTuning.ReservedCpuFreq = desiredReservedCoreFreq


josclark42:
@klaskosk @yliu127 Would you please review? Also, please let me know who on the system-tests side should be added as a reviewer as this PR affects their tests.

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -24,11 +25,18 @@ var _ = Describe("CPU frequency tuning tests change the core frequencies of isol
...
 			originalIsolatedCPUFreq performancev2.CPUfrequency
 			originalReservedCPUFreq performancev2.CPUfrequency
-			err                     error
 		)
 
 		BeforeEach(func() {
 
+			By("Checking that OCP version is 4.16 or later")
+			isGreaterOrEqual, err := rancluster.CompareOCPVersionWithCurrent(raninittools.Spoke1APIClient,
+				"4.18", true, false)
```
klaskosk, line 34: Should this be 4.16?

tests/internal/cluster/config.go
```diff
@@ -15,6 +16,77 @@ import (
...
+				return true, nil
+			}
+		} else {
+			if currentVersion.GreaterThan(referenceVersion) {
+				return true, nil
+			}
+		}
+	}
+
+	if orEqual {
```
klaskosk, line 75: I think this block also needs the `!isGreater` condition too. Might just be easier to add `return false, nil` at the end inside of the `if isGreater {}` block

tests/internal/cluster/config.go
```diff
@@ -15,6 +16,77 @@ import (
...
+		return false, fmt.Errorf("'referenceOCPVersion' cannot be empty")
+	}
+
+	currentOCPVersion, err := GetOCPVersion(apiClient)
+
+	if err != nil {
+		return false, err
+	}
+
+	glog.V(100).Infof("The apiClient is empty")
```
klaskosk, line 51:
```suggestion
	glog.V(90).Infof("The apiClient is empty")
```
Level 100 is reserved for eco-goinfra

klaskosk (changes requested):

tests/cnf/ran/powermanagement/tests/cpufreq.go
josclark42: Fixed- that was a left-over from debugging. Thanks

tests/internal/cluster/config.go
josclark42: Fixed- thanks.

tests/internal/cluster/config.go
josclark42: What do you think of this?

tests/internal/cluster/config.go
```diff
@@ -15,6 +16,77 @@ import (
...
+		if orEqual {
+			if currentVersion.LessThanOrEqual(referenceVersion) {
+				return true, nil
+			}
+		}
+	} else {
+		if currentVersion.LessThan(referenceVersion) {
+			return true, nil
+		}
+	}
```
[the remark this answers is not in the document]
josclark42: Yes- I made a mistake with the brackets. It should be fixed now, thanks.

tests/cnf/ran/powermanagement/tests/cpufreq.go
```diff
@@ -24,11 +25,18 @@ var _ = Describe("CPU frequency tuning tests change the core frequencies of isol
 			desiredIsolatedCoreFreq = performancev2.CPUfrequency(2200002)
 			originalIsolatedCPUFreq performancev2.CPUfrequency
 			originalReservedCPUFreq performancev2.CPUfrequency
-			err                     error
 		)
 
 		BeforeEach(func() {
 
+			By("Checking that OCP version is 4.16 or later")
+			isGreaterOrEqual, err := rancluster.CompareOCPVersionWithCurrent(raninittools.Spoke1APIClient,
```
yliu127, line 33: My suggestion is to do it the same way we do in TALM tests to be consistent within out own dir for any version checks: https://github.com/openshift-kni/eco-gotests/blob/abbd21ba1908f1d814e692abdff419fdff2c9280/tests/cnf/ran/talm/tests/talm_backup.go#L26 

@klaskosk I would suggest to initialize some common variables at ran level that are applicable to all or most of the ran test suites. And I think spoke version is one of those. And move the version comparison talm helper to ran helper. 

I think this approach is better because spoke version is static throughout the test, and there will probably be more scenarios like this in other test suites. Plus this way, we keep our changes under RAN dir. :)

klaskosk: @yliu127 I think the advantage of `CompareOCPVersionWithCurrent` is that it moves the version handling to another library that properly considers cases such as z-stream and `-rc` versions. Maybe we can add a different helper that does the actual version comparison so that we can use it separate from checking the OCP version? 

But yes I think it would be good to have a variable in `ranparam` for the OCP version that way we can compare throughout the test without checking again.

yliu127: @klaskosk our helper also handles zstreams, but yes using an existing library is nice. 

In this case, I would suggest do this in two steps: 
- ranparam change so Josh can use our helper 
- refactor our helper to use existing library

What do you think? The reason for refactoring our helper is because our desgin allows us to specify the start and the end releases in one call.

josclark42:
@yliu127 @klaskosk  I updated this PR based on the changes Kristen made in the ZTP PR to RanConfig. This makes version checks much simpler. I also added a check to the helper to create the performanceProfile if it does not exist which should fix the panic when the test is run in CI.

klaskosk (approved):
/lgtm

PR #56, opened 2024-06-20
lca: cnf exclude operators and sriov nnp from post validation checks

mcornea:
- exclude operators whose version does not change during upgrade
- exclude default sriov nnp which is not present in 4.16
- exclude null characters when validating seed entries in etcd for a cleaner output


yliu127 (approved):

mpmaruthu:
Looks good to me.
_Peri M_

PR #58, opened 2024-06-20
lca: cnf retry kcat command execution

mcornea:
kcat command may fail so retrying multiple times until the command exist without errors.


mpmaruthu:
Looks good to me.
/Peri M

PR #61, opened 2024-06-21
Add stability tests for ran du

rlmanrique:
This PR adds two new test cases for stability in RAN env:

- Stability without workload runnning
- Stability with workload runing
-

Both tests collect a series of metrics and verify changes on them at the end of the test duration.

It also includes several functions to be used by other tests such as PTP validators and stability helpers.


rlmanrique:
Both tests are missing the report ID XML. I need to check internally what values could be set. Apart from that, the PR is ready for review.

tests/system-tests/ran-du/tests/stability-workload.go
```diff
@@ -0,0 +1,136 @@
...
+			Expect(err).ToNot(HaveOccurred(), "pod not ready: %s", err)
+
+			By("Fetching Cluster name")
+			clusterName, err = platform.GetOCPClusterName(APIClient)
+			Expect(err).ToNot(HaveOccurred(), "Failed to get cluster name")
+
+		})
+		It("StabilityWorkload", reportxml.ID("TBD"), Label("StabilityWorkload"), func() {
+
+			outputDir := RanDuTestConfig.StabilityOutputPath
```
rlmanrique, line 63: Is there a convention on which output path can be used to store output files?

tests/system-tests/internal/ptp/ptp.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/pod"
+)
+
+func isPtPClockSync(apiClient *clients.Settings) (bool, error) {
```
mcornea, line 12: afaik timedatectl reports the status based on the NTP sync. For checking the PTP clock sync I think a better approach is to query the openshift_ptp_clock_state metric, e.g.

```
oc -n openshift-ptp exec linuxptp-daemon-h4kk4 -- curl -s http://localhost:9091/metrics | grep openshift_ptp_clock_state
Defaulted container "cloud-event-proxy" out of: cloud-event-proxy, kube-rbac-proxy, linuxptp-daemon-container
# HELP openshift_ptp_clock_state 0 = FREERUN, 1 = LOCKED, 2 = HOLDOVER
# TYPE openshift_ptp_clock_state gauge
openshift_ptp_clock_state{iface="CLOCK_REALTIME",node="sno.kni-qe-24.lab.eng.rdu2.redhat.com",process="phc2sys"} 1
openshift_ptp_clock_state{iface="ens1fx",node="sno.kni-qe-24.lab.eng.rdu2.redhat.com",process="ptp4l"} 1
```

tests/system-tests/ran-du/tests/stability-no-workload.go
```diff
@@ -0,0 +1,106 @@
...
+			if RanDuTestConfig.PtpEnabled {
+				_, err = stability.VerifyStabilityStatusChange(ptpOutputFile)
+				if err != nil {
+					stabilityErrors = append(stabilityErrors, err.Error())
+				}
+			}
+
+			By("Check if there been any error")
+			if len(stabilityErrors) > 0 {
+				Expect(stabilityErrors).ToNot(HaveOccurred(), "One or more errors in stability tests")
```
mcornea, line 100: Can we also print the stabilityErrors to make it easier to understand the errors in case of failure?

tests/system-tests/ran-du/tests/stability-workload.go
mcornea: We don't have any convention but I think `/tmp` is good enough for this.

tests/system-tests/internal/ptp/ptp.go
rlmanrique: Great. I didn't know about it. 

I added it as an extra check. Thansk for the suggestion. Let me know if current approach is good enough.

tests/system-tests/ran-du/tests/stability-no-workload.go
rlmanrique: Sure, done.

tests/system-tests/ran-du/tests/stability-workload.go
rlmanrique: I finally set it in by default in /tmp/resports

mcornea:
lgtm

mcornea (approved):

PR #62, opened 2024-06-21
ibu mgmt: add missing dpa test

trewest:
Add missing DPA test to the negative test suite. Also update IBU clean up to be contained in one suite level AfterEach container.

### DIFF
--- a/tests/lca/imagebasedupgrade/mgmt/negative/negative_suite_test.go
+++ b/tests/lca/imagebasedupgrade/mgmt/negative/negative_suite_test.go
@@ -3,16 +3,19 @@ package negative_test
 import (
 	"runtime"
 	"testing"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/lca"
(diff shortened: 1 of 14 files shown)


tests/lca/imagebasedupgrade/mgmt/negative/negative_suite_test.go
```diff
@@ -35,6 +38,48 @@ var _ = BeforeSuite(func() {
...
+		Expect(err).NotTo(HaveOccurred(), "error waiting for idle stage to complete")
+	}
+
+	Eventually(func() (bool, error) {
+		ibu.Object, err = ibu.Get()
+		if err != nil {
+			return false, err
+		}
+
+		return len(ibu.Object.Status.Conditions) == 1 &&
```
achuzhoy, line 74: why do we check for having a single condition?

trewest: Because when the IBU is in a proper Idle state, there is only one condition. e.g.
```
$ oc get ibu upgrade -o yaml | yq e .status -  
conditions:
  - lastTransitionTime: "2024-06-21T15:28:39Z"
    message: Idle
    observedGeneration: 197
    reason: Idle
    status: "True"
    type: Idle
observedGeneration: 197
validNextStages:
  - Prep
```

achuzhoy (approved):

PR #65, opened 2024-06-24
RDS Core: separate general methods from the sriov verification suite

elenagerman:
- helper.go file created

- verifySRIOVConnectivity was renamed to the verifyConnectivity and moved to the helper.go

- createServiceAccount, deleteServiceAccount, deleteClusterRBAC, createClusterRBAC, deleteConfigMap, createConfigMap, deleteDeployments, findPodWithSelector, defineContainer, verifyMsgInPodLogs were moved to the helper.go

- 'if pState == "NonCompliant"' state was changed to the 'if pState == "NonCompliant"'. Depends on deployment type it can be some extra policies with no state at all be found and it can fail test with no reason


elenagerman:
@yprokule Hi! as was discussed, PTAL

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,507 @@
...
+)
+
+func createServiceAccount(saName, nsName string) {
+	ginkgo.By(fmt.Sprintf("Creating ServiceAccount %q in %q namespace",
+		saName, nsName))
+
+	glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Creating SA %q in %q namespace",
+		saName, nsName)
+
+	insureNamespaceExists(nsName)
```
yprokule, line 32: this shouldn't be part of the method, it should only create SA, _namespace_ management should be done outside

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,507 @@
...
+		}).WithContext(ctx).WithPolling(5*time.Second).WithTimeout(1*time.Minute).Should(gomega.BeTrue(),
+			"Failed to delete configMap")
+	}
+}
+
+func createConfigMap(cmName, nsName string, data map[string]string) {
+	glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Create ConfigMap %q in %q namespace",
+		cmName, nsName)
+
+	insureNamespaceExists(nsName)
```
yprokule, line 217: ditto

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,507 @@
...
+		glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Logs from pod %s:\n%s", podObj.Definition.Name, podLog)
+
+		return true
+	}).WithContext(ctx).WithPolling(5*time.Second).WithTimeout(1*time.Minute).Should(gomega.BeTrue(),
+		fmt.Sprintf("Failed to get logs from pod %q", podObj.Definition.Name))
+
+	gomega.Expect(podLog).Should(gomega.ContainSubstring(msg))
+}
+
+func insureNamespaceExists(nsName string) {
```
yprokule, line 414: I assume it should be `ensure` and not `insure` 😸

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,471 @@
...
+
+	if !deploySa.Exists() {
+		var ctx ginkgo.SpecContext
+
+		deploySa, err := deploySa.Create()
+
+		if err != nil {
+			glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Error creating SA %q in %q namespace: %v",
+				saName, nsName, err)
+		}
```
yprokule, line 41: This should be inside `Eventually` block

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,471 @@
...
+func createServiceAccount(saName, nsName string) {
+	ginkgo.By(fmt.Sprintf("Creating ServiceAccount %q in %q namespace",
+		saName, nsName))
+
+	glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Creating SA %q in %q namespace",
+		saName, nsName)
+
+	deploySa := serviceaccount.NewBuilder(rdscoreinittools.APIClient, saName, nsName)
+
+	if !deploySa.Exists() {
```
yprokule, line 33: please add logging for a case that `SA` already exists, thanks

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,471 @@
...
+			saName, nsName)
+		glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Deleting ServiceAccount %q in %q namespace",
+			saName, nsName)
+
+		err := deploySa.Delete()
+
+		if err != nil {
+			glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Error deleting ServiceAccount %q in %q namespace: %v",
+				saName, nsName, err)
+		}
```
yprokule, line 79: This should be within `Eventually` block

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,471 @@
...
+		rdscoreinittools.APIClient,
+		rbacName); err == nil {
+		glog.V(rdscoreparams.RDSCoreLogLevel).Infof("ClusterRoleBinding %q found. Deleting...", rbacName)
+
+		err := crbSa.Delete()
+
+		if err != nil {
+			glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Error deleting ClusterRoleBinding %q : %v",
+				rbacName, err)
+		}
```
yprokule, line 118: ditto, should be within `Eventually` block

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,471 @@
...
+			Name:      saName,
+			Kind:      "ServiceAccount",
+			Namespace: nsName,
+		})
+
+	crbSa, err := crbSa.Create()
+	if err != nil {
+		glog.V(rdscoreparams.RDSCoreLogLevel).Infof(
+			"Error Creating ClusterRoleBinding %q : %v", crbSa.Definition.Name, err)
+	}
```
yprokule, line 157: ditto, should be within `Eventually` block

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,471 @@
...
+		glog.V(rdscoreparams.RDSCoreLogLevel).Infof("configMap %q found, deleting", cmName)
+
+		var ctx ginkgo.SpecContext
+
+		err := cmBuilder.Delete()
+
+		if err != nil {
+			glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Error deleting configMap %q : %v",
+				cmName, err)
+		}
```
yprokule, line 190: ditto

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,471 @@
...
+	cmBuilder.WithData(data)
+
+	var ctx ginkgo.SpecContext
+
+	cmResult, err := cmBuilder.Create()
+	if err != nil {
+		glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Error creating ConfigMap %q in %q namespace: %v",
+			cmName, nsName, err)
+	}
+
```
yprokule, line 222: ditto

tests/system-tests/rdscore/internal/rdscorecommon/helper.go
```diff
@@ -0,0 +1,471 @@
...
+}
+
+func deleteDeployments(dName, nsName string) {
+	ginkgo.By(fmt.Sprintf("Removing test deployment %q from %q ns", dName, nsName))
+
+	if deploy, err := deployment.Pull(rdscoreinittools.APIClient, dName, nsName); err == nil {
+		glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Deleting deployment %q from %q namespace",
+			deploy.Definition.Name, nsName)
+
+		deploy, err = deploy.WithReplicas(int32(0)).Update()
```
yprokule, line 247: no need to scale down deployment before deletion

yprokule (changes requested):
please address new comments, thanks

elenagerman:
@yprokule I'm sorry, but I do not understand why general methods like delete/create API objects cannot be outside the sriov validation test. What if we need it for the other operator's tests? For example, configmap, rbac, and sa config can be required for the keda config.
I also do not understand what is terrible in double-checking if ns exists.
Anyway, with all the requested changes, we are returning to the original version; I do not see any value in this MR.
Thank you!

yprokule:
> @yprokule I'm sorry, but I do not understand why general methods like delete/create API objects cannot be outside the sriov validation test. What if we need it for the other operator's tests? For example, configmap, rbac, and sa config can be required for the keda config.

@elenagerman what are U talking about? I didn't object about having some of the common methods in a dedicated file. What I said is that during your _move_ of existing methods to a new file U broke the existing logic and were asked to fix it

> I also do not understand what is terrible in double-checking if ns exists.

U don't double check but create if it's missing which again affects existing logic. And what the problem to run this at the beginning of the test or in the setup block?

> Anyway, with all the requested changes, we are returning to the original version; I do not see any value in this MR.

Again, not sure where U got the idea that we are returning to the original version - all U had to do was to address broken logic in the PR.

Feel free to reopen PR and address comments or I might piggyback on this work and adjust it myself.

Thank you!

PR #69, opened 2024-06-26
vCore: numa test suite

elenagerman:
vCore: numa test suite

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/nmstate/kubernetes-nmstate/api v0.0.0-20240605150941-df565dd7bf35
 	github.com/onsi/ginkgo/v2 v2.19.0
 	github.com/onsi/gomega v1.33.1
-	github.com/openshift-kni/eco-goinfra v0.0.0-20240621192821-7c55fda74f32 // release-4.15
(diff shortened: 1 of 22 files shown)


go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/nmstate/kubernetes-nmstate/api v0.0.0-20240605150941-df565dd7bf35
 	github.com/onsi/ginkgo/v2 v2.19.0
 	github.com/onsi/gomega v1.33.1
-	github.com/openshift-kni/eco-goinfra v0.0.0-20240621192821-7c55fda74f32 // release-4.15
+	github.com/openshift-kni/eco-goinfra v0.0.0-20240626143745-4d019aeb6f02 // release-4.15
```
yprokule, line 21: this should be done outside of this PR

elenagerman: why?

elenagerman: @yprokule with no bumping eco-goinfra I cannot use new method

yprokule: @elenagerman because there is [workflow](https://github.com/openshift-kni/eco-gotests?tab=readme-ov-file#update-eco-goinfra-modules---how-to) - U update `eco-goinfra` in a separate PR then rebase your changes on top of it.

PR #70, opened 2024-06-27
lca: cnf: Retry seed image info retrieval in post validations

mcornea:
Sometimes retrieving the seed image info may fail so re-trying until no errors are reported.


mpmaruthu:
Looks good to me.
/Peri M

PR #72, opened 2024-06-27
cnf ran: ztp test migration preliminaries

klaskosk:
This is the first PR for migrating the ZTP test suite and includes the ztp directory with the helper and tsparams packages. Additionally, some helpers and parameters are moved from the TALM suite into the ran directory so they may be shared with the ZTP tests.


klaskosk:
Verified TALM suite passes after these changes

tests/cnf/ran/internal/ranhelper/version.go
```diff
@@ -0,0 +1,211 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// InitializeSpokeNames initializes the name of spoke 1 and, if present, spoke 2.
+func InitializeSpokeNames() error {
+	var err error
+
+	// Spoke 1 is required to be present.
+	ranparam.Spoke1Name, err = getClusterName(os.Getenv("KUBECONFIG"))
```
sebrandon1, line 26: Just as an overall comment, not necessarily something wrong with this functionality of this PR because I think this "works" fine but global variables make things near impossible to trace and to unit test if needed in the future.  These funcs that set `ranparam` values should be using parameters and return values.

It's probably fine for this PR since this is pretty much repo-wide at this point but I would love to see a better way of doing this in the future.

tests/cnf/ran/internal/ranhelper/version.go
```diff
@@ -0,0 +1,211 @@
...
+	ranparam.ZtpVersion, err = getZtpVersionFromArgoCd(
+		raninittools.HubAPIClient, ranparam.OpenshiftGitopsRepoServer, ranparam.OpenshiftGitops)
+
+	return err
+}
+
+// IsVersionStringInRange checks if a version string is between a specified min and max value, inclusive. All the string
+// inputs to this function should be dot separated positive intergers, e.g. "1.0.0" or "4.10". Each string input must be
+// at least two dot separarted integers but may also be 3 or more, though only the first two are compared.
+func IsVersionStringInRange(version, minimum, maximum string) (bool, error) {
```
sebrandon1, line 76: This seems like a good candidate for unit tests to test functionality.

tests/cnf/ran/internal/ranhelper/version.go
```diff
@@ -0,0 +1,211 @@
...
+			return false, nil
+		}
+	}
+
+	// At the end if we never returned then all the digits were in valid range
+	return true, nil
+}
+
+// validateInputString validates that a string is at least two dot separated nonnegative integers.
+func validateInputString(input string) (bool, []int) {
```
sebrandon1, line 125: Same here, a unit test to lock in the functionality would be good for this one.

sebrandon1: This could also be a regex to validate the structure of the string.

Something like `regexp.MatchString(input, "[0-9]+.[0-9]+")`.

sebrandon1 (commented):
Some comments.

tests/cnf/ran/internal/ranhelper/version.go
klaskosk: I have to agree that the global mutable state is pretty non ideal. For this PR I'll go ahead and implement Yang's suggestion which will centralize the global state into the config. This should organize things a bit more if we ever do want to make an effort to move away from global variables everywhere

klaskosk:
@yliu127 @sebrandon1 Thanks for the comments, I've gone ahead and updated the PR and verified TALM still works again

tests/cnf/ran/internal/ranconfig/config.go
```diff
@@ -20,20 +25,49 @@ const (
...
+	HubAPIClient  *clients.Settings
+	HubOCPVersion string
+	ACMVersion    string
+	TALMVersion   string
+	ZTPVersion    string
+	HubKubeconfig string `envconfig:"ECO_CNF_RAN_KUBECONFIG_HUB"`
+}
+
+// Spoke1Config contains the configuration for the spoke 1 cluster, which should always be present.
+type Spoke1Config struct {
```
yliu127, line 54: I think we can use only one struct and one constructor func newSpokeConfig that returns SpokeConfig? e.g., type SpokeConfig struct {...}, then var Spoke1Config SpokeConfig . 

func newSpokeConfig (configFile string) SpokeConfig {...}
If the config file does not contain certain field, then just set them to empty/nil.   

Spoke1Config = newSpokeConfig(spoke1ConfigFile)
Spoke2Config = newSpokeConfig(spoke2ConfigFile)  # if spoke2 exists

And for SpokeConfig and HubConfig struct, I think we can combine the day2 operator versions into one map field. So it is easier to get all relevant operator versions in one function call.

e.g., 
```
type HubConfig struct {
	APIClient  *clients.Settings
	OCPVersion string
	Day2OperatorVersions map[string]string
        ZTPVersion string
	Kubeconfig string `envconfig:"ECO_CNF_RAN_KUBECONFIG_HUB"`
}

type SppkeConfig struct {
	BMC        *bmc.BMC
	APIClient  *clients.Settings
	ClusterName       string
	OCPVersion string
        Day2OperatorVersions map[string]string
        Kubeconfig string
	BMCUsername      string 
        BMCPassword      string 
	BMCHosts         []string 
	BMCTimeout       time.Duration `yaml:"bmcTimeout" envconfig:"ECO_CNF_RAN_BMC_TIMEOUT"`
}

```

func GetOperatorVersionsFromCsv(client *clients.Settings, operatorNames map[string]string) map[string]string {...}

operatorNames uses namespace as key and name as value.
return value can be a map with operator name as key and version as value.

Alternatively, OperatorInfo could be a struct itself, and both input can be a list of OperatorInfo . I think this method will require more code and not sure if it will give us much benefit since at this top level we will very likely never need to collect/store more info other than name, namespace and version for an operator. 

For spoke operators, we will need to collect ptp, sriov, logging, local-storage, and lifecycle-agent versions. And LCA might not be installed in all envs.

Let me know how you think about this.

tests/cnf/ran/ztp/ztp_suite_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "RAN ZTP Suite", Label(tsparams.Labels...), reporterConfig)
+}
+
+var _ = BeforeSuite(func() {
+	err := helper.GetArgoCdAppGitDetails()
+	Expect(err).ToNot(HaveOccurred(), "Failed to get current data from ArgoCD")
+
+	By("deleting and recreating ZTP test namespace to ensure a blank state")
```
mpmaruthu, line 33: I would approach first with Exists() check and then delete ztp test namespace if exists

tests/cnf/ran/ztp/ztp_suite_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	_, err = namespace.NewBuilder(HubAPIClient, tsparams.TestNamespace).Create()
+	Expect(err).ToNot(HaveOccurred(), "Failed to create ZTP test namespace")
+})
+
+var _ = AfterSuite(func() {
+	// err := helper.ResetArgoCdGitDetails()
+	// Expect(err).ToNot(HaveOccurred(), "Failed to restore original ArgoCD configuration")
+
+	err := namespace.NewBuilder(HubAPIClient, tsparams.TestNamespace).DeleteAndWait(5 * time.Minute)
```
mpmaruthu, line 45: I would approach first with Exists() check and then delete ztp test namespace if exists

mpmaruthu (commented):
2 minor comments.

tests/cnf/ran/internal/ranconfig/config.go
klaskosk: @yliu127 I considered using a single SpokeConfig struct before but ultimately decided against it since it prevents using envconfig with struct tags properly. Instead, we would have to read (and parse) different environment variables depending on which spoke config is being loaded. Also, Spoke 2 is only used by TALM and never for BMC so it doesn't seem unreasonable to consider it a different type unless it starts being used more similarly to Spoke 1.

I like the idea for collecting spoke operator versions, though we won't need them until the PTP suite gets migrated so maybe we can add them to the config during that suite?

tests/cnf/ran/ztp/ztp_suite_test.go
klaskosk: @mpmaruthu Namespace is one of the resources where [deleting returns nil instead of an error if it does not exist](https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/namespace/namespace.go#L158) so it shouldn't cause errors in this case

tests/cnf/ran/ztp/ztp_suite_test.go
mpmaruthu: In this case, my comment become invalid.  Please ignore comments.Thanks!

tests/cnf/ran/internal/ranconfig/config.go
yliu127: @klaskosk thanks for clarifying. Let's just do the day2 operator change for hub first? We need acm, mce, gitops, and talm for hub. Although in the ztp/talm suite, we only need to use ztp/talm version to determine whether a test is valid, but when test fails, we often need to provide versions for all relevant components.

yliu127 (approved):

tests/cnf/ran/internal/ranconfig/config.go
```diff
@@ -47,24 +80,158 @@ func NewRANConfig() *RANConfig {
...
+	ranconfig.Spoke1Config = new(Spoke1Config)
+
+	err := readConfig(ranconfig.Spoke1Config, configFile)
+	if err != nil {
+		glog.V(ranparam.LogLevel).Infof("Failed to instantiate Spoke1Config: %v", err)
+	}
+
+	ranconfig.Spoke1Config.Spoke1APIClient = inittools.APIClient
+
+	spoke1Kubeconfig := os.Getenv("KUBECONFIG")
```
sebrandon1, line 169: Should this have a safety check in-case `KUBECONFIG` isn't available?

tests/cnf/ran/internal/ranhelper/version.go
```diff
@@ -0,0 +1,180 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/deployment"
+	"github.com/openshift-kni/eco-goinfra/pkg/olm"
+	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/ranparam"
+	"github.com/openshift-kni/eco-gotests/tests/internal/cluster"
+	v1 "github.com/openshift/api/config/v1"
```
sebrandon1, line 17:
```suggestion
	configv1 "github.com/openshift/api/config/v1"
```

sebrandon1 (approved):
Looks good.  Just a couple of comments.

PR #74, opened 2024-06-28
system-test: Unify delete workload mechanism in ran du tests

rlmanrique:


tests/system-tests/ran-du/tests/stability-no-workload.go
```diff
@@ -23,6 +24,10 @@ var _ = Describe(
 		)
 		BeforeAll(func() {
 
+			By("Cleaning up test workload resources")
```
mcornea, line 27: If the workload is not running, then the delete script may fail. Therefore, I think we should condition it on the existence of the namespace.": 

```suggestion
			if namespace.NewBuilder(APIClient, RanDuTestConfig.TestWorkload.Namespace).Exists() {
				By("Deleting workload using shell method")
				_, err := shell.ExecuteCmd(RanDuTestConfig.TestWorkload.DeleteShellCmd)
				Expect(err).ToNot(HaveOccurred(), "Failed to delete workload")
			}
```

rlmanrique: Good catch! Check added

mcornea:
lgtm

PR #75, opened 2024-06-28
system-test: Enable ptp checks in existing test cases

rlmanrique:
Also, include small improvements like customize reboot
recovery time. Also, fixes the stability output file
between two different tests


mcornea:
lgtm

PR #79, opened 2024-07-04
lca: cnf: wait for ibu idle stage after finalize cgu

mcornea:
After the finalize CGU reports Completed state wait for the IBU Idle condition to report True as well. Also remove 'Wait until node is reporting as Ready' as it can sometimes prematurely fail if the kube api is not available.  In addition wait 10 seconds after the CGUs deletion as it seems that if we delete and re-create the CGUs too quickly the policies do not have enough time to reconciliate their status.


mpmaruthu:
Looks good to me.
/Peri M

mpmaruthu:
few nit comments from my side otherwise looks good to me.
/Peri M

mpmaruthu:
Looks good to me.
/Peri M

mpmaruthu (approved):

PR #82, opened 2024-07-08
system-tests: ran-du: add tuned restarts check to stability tests

mcornea:
In order to catch issues such as [OCPBUGS-30813 ](https://issues.redhat.com/browse/OCPBUGS-30813)add a check to the stability test which validates that tuned does not continuously apply its configuration.


rlmanrique (approved):
LGTM

PR #90, opened 2024-07-10
cnf ran: add reporting for CRs on each cluster

klaskosk:
Currently, the cnf/ran test suites do not log the CRs used in each test case, which becomes particularly important for the TALM and ZTP test suites. This PR uses the existing infrastructure in the repo around reporting to generate dumps with these resources.

There is an additional function in the internal/reporter package that allows reports to dump the CRs from a specific kubeconfig, not just the one specified by the KUBECONFIG environment variable.

### DIFF
--- a/tests/cnf/ran/talm/internal/tsparams/talmvars.go
+++ b/tests/cnf/ran/talm/internal/tsparams/talmvars.go
@@ -3,23 +3,51 @@ package tsparams
 import (
 	"fmt"
 
+	cguv1alpha1 "github.com/openshift-kni/cluster-group-upgrades-operator/pkg/api/clustergroupupgrades/v1alpha1"
 	"github.com/openshift-kni/eco-gotests/tests/cnf/ran/internal/ranparam"
 	"github.com/openshift-kni/k8sreporter"
-	v1 "k8s.io/api/core/v1"
+	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	policiesv1beta1 "open-cluster-management.io/governance-policy-propagator/api/v1beta1"
+	placementrulev1 "open-cluster-management.io/multicloud-operators-subscription/pkg/apis/apps/placementrule/v1"
 )
 
 var (
 	// Labels represents the range of labels that can be used for test cases selection.
 	Labels = append(ranparam.Labels, LabelSuite)
 
-	// ReporterNamespacesToDump tells to the reporter from where to collect logs.
-	ReporterNamespacesToDump = map[string]string{
-		TestNamespace: "",
+	// ReporterHubNamespacesToDump tells to the reporter which namespaces on the hub to collect pod logs from.
+	ReporterHubNamespacesToDump = map[string]string{
+		TestNamespace:                       "",
+		ranparam.OpenshiftOperatorNamespace: "",
 	}
 
-	// ReporterCRDsToDump tells to the reporter what CRs to dump.
-	ReporterCRDsToDump = []k8sreporter.CRData{
-		{Cr: &v1.PodList{}},
+	// ReporterSpokeNamespacesToDump tells the reporter which namespaces on the spokes to collect pod logs from.
+	ReporterSpokeNamespacesToDump = map[string]string{
+		TemporaryNamespace:  "",
+		PreCacheSpokeNS:     "",
+		TalmBackupNamespace: "",
+	}
+
+	// ReporterHubCRsToDump is the CRs the reporter should dump on the hub.
+	ReporterHubCRsToDump = []k8sreporter.CRData{
+		{Cr: &corev1.NamespaceList{}},
+		{Cr: &corev1.PodList{}},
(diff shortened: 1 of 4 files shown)


tests/cnf/ran/talm/internal/tsparams/talmvars.go
```diff
@@ -14,12 +20,21 @@ var (
...
+		{Cr: &corev1.PodList{}},
+		{Cr: &corev1.NamespaceList{}},
+		{Cr: &cguv1alpha1.ClusterGroupUpgradeList{}, Namespace: ptr.To(TestNamespace)},
+		{Cr: &cguv1alpha1.PreCachingConfigList{}, Namespace: ptr.To(TestNamespace)},
+		{Cr: &policiesv1.PolicyList{}, Namespace: ptr.To(TestNamespace)},
+		{Cr: &policiesv1.PlacementBindingList{}, Namespace: ptr.To(TestNamespace)},
+		{Cr: &placementrulev1.PlacementRuleList{}, Namespace: ptr.To(TestNamespace)},
+		{Cr: &policiesv1beta1.PolicySetList{}, Namespace: ptr.To(TestNamespace)},
+		{Cr: &olmv1alpha1.CatalogSourceList{}, Namespace: ptr.To(TemporaryNamespace)},
 	}
```
yliu127, line 38: We will need a  unique list for hub and spoke.  And besides CR, is there a way to collect pod logs?
Also I assume any "List" actual collects individual CRs in yaml format? i.e., CGU list will allow us to check the specific cgu status, etc. 

For spoke, we need the following: 
- pod list
- namespace list
- precache/backup job(s)  (I'm not sure what is the NS, should probably check existing bug report or TALM repo or ask Jun)
- precache/backup pod log(s) (same here re NS)

tests/cnf/ran/talm/talm_suite_test.go
```diff
@@ -42,9 +44,36 @@ var _ = AfterSuite(func() {
...
+		reporter.ReportIfFailedOnCluster(
+			RANConfig.HubKubeconfig,
+			report,
+			hubReportPath,
+			tsparams.ReporterNamespacesToDump,
+			tsparams.ReporterCRDsToDump,
+			clients.SetScheme)
+	}
+
+	if Spoke2APIClient != nil {
```
yliu127, line 68: what about spoke1?

klaskosk: Spoke 1 is in KUBECONFIG so it's just the default one returned. Only Spoke 2 and Hub need to have explicit kubeconfigs set

tests/cnf/ran/talm/internal/tsparams/talmvars.go
klaskosk: @yliu127 I've kept all the CRs in the same list here since it's not an error to try to list CRs that don't exist. But I can separate them into a list per hub/spoke if that's clearer. Same with the namespaces for pod logs.

But yes the list resource is just the way the reporter wants it and means we want to list all resources of that kind.

It uses the `ReporterNamespacesToDump` variable to determine which pod logs to dump. I'll have to update that to collect more relevant logs though since the namespaces we create won't have any pods.

yliu127: I think separate list is better because hub and spoke CRs are pretty much all different. 

Oh I see, in that case we should use cgu pod NS (openshift-operators) for pod log on hub. And for spoke, we need to find out which NS is used for precache/backup pods.

tests/cnf/ran/talm/talm_suite_test.go
yliu127: Oh so I guess spoke1 is covered by line 56.

tests/cnf/ran/talm/internal/tsparams/talmvars.go
klaskosk: Okay, I'll go ahead and separate them. The precache pod is already in the tsparams so I'll just use that namespace. I'll have to look for the backup pod though since it's not in the test

tests/cnf/ran/talm/internal/tsparams/talmvars.go
```diff
@@ -14,12 +20,21 @@ var (
 
 	// ReporterNamespacesToDump tells to the reporter from where to collect logs.
 	ReporterNamespacesToDump = map[string]string{
-		TestNamespace: "",
+		TestNamespace:      "",
+		TemporaryNamespace: "",
 	}
 
 	// ReporterCRDsToDump tells to the reporter what CRs to dump.
 	ReporterCRDsToDump = []k8sreporter.CRData{
```
yliu127, line 28: Let's use this name for spoke since spoke kubeconfig is the default, and then a ReporterHubCRsToDump for hub? 

Btw, it should be CR to dump instead of CRD as CRD is a static definition of what fields a CR can use. CR is the actual config applied.

yliu127: Or maybe D here meant data? -_-

klaskosk: @yliu127 I'm not sure since I think I just copied it over during the migration. But either way CR seems more precise and it's what the comment uses anyway

klaskosk:
@yliu127 I've updated the lists of namespaces and CRs now to align with we need on the hub and spokes

yliu127 (approved):
/lgtm

kononovn (approved):
/lgtm

PR #108, opened 2024-07-19
cnf ran: add secret creation to ztp hub templating

klaskosk:
This adds the secret creation to the code for the hub templating test rather than letting it be created by Argo CD since that is not permitted.

It additionally fixes an issue that has occured with the image registry object being nil.


yliu127 (approved):
/lgtm

sebrandon1 (approved):

PR #119, opened 2024-07-25
cnf ran: ztp reporting and improvements

klaskosk:
1. Reporting for the custom resources has been added to the ZTP test suite. The caveat is that until eco-goinfra support is added it will be missing the Argo CD Applications CR.
2. Image registry tests have been changed a bit to test for the path in git before failing if /var/imageregistry is not present.
3. An additional log has been added to the hub templating tests to provide more information for a failure that ocurred in CI.


yliu127 (approved):
/lgtm

josclark42 (approved):

PR #121, opened 2024-07-25
added version check to test 54240

josclark42:
Test 54240 requires ZTP 4.15 or less. Added a check to this test.


tests/cnf/ran/ztp/tests/ztp-argocd-hub-templating.go
```diff
@@ -65,11 +65,19 @@ var _ = Describe("ZTP Argo CD Hub Templating Tests", Label(tsparams.LabelArgoCdH
 	It("should report an error for using autoindent function where not allowed", reportxml.ID("54240"), func() {
 		setupHubTemplateTest(tsparams.ZtpTestPathTemplatingAutoIndent)
```
klaskosk, line 66: We should check for the version before we get the test set up so that we're not updating argo cd, creating a cgu, etc. if we don't need it

josclark42: Fixed. Thank you.

klaskosk (approved):

kononovn (approved):

klaskosk:
@josclark42 This ended up not needing a version check since the issue is autoindent being used with an integer argument, which causes an error (as desired) even in ZTP 4.16 and above. Just closing this now so it's not lingering but happy to revisit later if the situation changes

PR #125, opened 2024-07-29
cnf ran: image registry CR report and spoke checker fix

klaskosk:
1. The list of spoke CRs to report for the ZTP suite test is currently missing the image registry config despite this CR being used in the policies app test.
2. The spoke checker test currently assumes that the cluster is SNO which causes it to fail for SNO+1. This PR updates it to perform the check on each node.

### DIFF
--- a/tests/cnf/ran/ztp/tests/ztp-spoke-checker.go
+++ b/tests/cnf/ran/ztp/tests/ztp-spoke-checker.go
@@ -20,26 +20,31 @@ var _ = Describe("ZTP Spoke Checker Tests", Label(tsparams.LabelSpokeCheckerTest
 			By("verifying chronyd is inactive")
 			// Use `| cat -` to explicitly ignore the return code since it will be nonzero when the service
 			// is not active, which is expected here.
-			status, err := cluster.ExecCommandOnSNO(
-				Spoke1APIClient, 3, "systemctl is-active chronyd.service | cat -")
+			statuses, err := cluster.ExecCmdWithStdout(Spoke1APIClient, 3, "systemctl is-active chronyd.service | cat -")
 			Expect(err).ToNot(HaveOccurred(), "Failed to check if chronyd service is active")
+			Expect(statuses).ToNot(BeEmpty(), "Failed to find statuses for chronyd service")
 
-			glog.V(tsparams.LogLevel).Infof("active status: %s", status)
+			for nodeName, status := range statuses {
+				glog.V(tsparams.LogLevel).Infof("%s active status: %s", nodeName, status)
 
-			status = strings.TrimSpace(status)
-			Expect(status).To(Equal("inactive"), "chronyd service was not inactive")
+				status = strings.TrimSpace(status)
+				Expect(status).To(Equal("inactive"), "chronyd service was not inactive")
+			}
 
 			By("verifying chronyd is disabled")
 			// Use `| cat -` to explicitly ignore the return code since it will be nonzero when the service
 			// is disabled, which is expected here.
-			status, err = cluster.ExecCommandOnSNO(
-				Spoke1APIClient, 3, "systemctl is-enabled chronyd.service | cat -")
+			statuses, err = cluster.ExecCmdWithStdout(Spoke1APIClient, 3, "systemctl is-enabled chronyd.service | cat -")
 			Expect(err).ToNot(HaveOccurred(), "Failed to check if chronyd service is enabled")
+			Expect(statuses).ToNot(BeEmpty(), "Failed to find statuses for chronyd service")
 
-			glog.V(tsparams.LogLevel).Infof("enabled status: %s", status)
+			for nodeName, status := range statuses {
+				glog.V(tsparams.LogLevel).Infof("%s enabled status: %s", nodeName, status)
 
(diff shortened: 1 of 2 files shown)


klaskosk:
Tested these changes against the SNO worker environment successfully

tests/cnf/ran/ztp/tests/ztp-spoke-checker.go
```diff
@@ -6,40 +6,66 @@ import (
...
 var _ = Describe("ZTP Spoke Checker Tests", Label(tsparams.LabelSpokeCheckerTests), func() {
+	var nodeSelector metav1.ListOptions
+
+	BeforeEach(func() {
+		nodeList, err := nodes.List(Spoke1APIClient)
+		Expect(err).ToNot(HaveOccurred(), "Failed to list nodes for spoke 1")
+		Expect(nodeList).ToNot(BeEmpty(), "Failed to find any node for spoke 1")
+
+		nodeSelector = metav1.ListOptions{
+			FieldSelector: fields.SelectorFromSet(fields.Set{"metadata.name": nodeList[0].Definition.Name}).String(),
```
yliu127, line 28: Perhaps we should check all nodes instead of the first node. As in the case of sno+worker, the two nodes are not equal, they got installed differently, so it would be nice to check both with a loop inside the test function?

tests/cnf/ran/ztp/tests/ztp-spoke-checker.go
```diff
@@ -6,40 +6,66 @@ import (
...
 	When("a TunedPerformancePatch.yaml PGT disables the chronyd service", func() {
 		// 54237 - Moving chronyd disable to tuned patch
 		It("should disable and inactivate the chronyd service on spoke", reportxml.ID("54237"), func() {
 			By("verifying chronyd is inactive")
 			// Use `| cat -` to explicitly ignore the return code since it will be nonzero when the service
 			// is not active, which is expected here.
-			status, err := cluster.ExecCommandOnSNO(
-				Spoke1APIClient, 3, "systemctl is-active chronyd.service | cat -")
+			statuses, err := cluster.ExecCmdWithStdout(
+				Spoke1APIClient, 3, "systemctl is-active chronyd.service | cat -", nodeSelector)
```
yliu127, line 39: We can probably select all the nodes with worker role. So it doesn't need additional loop other than defining the nodeSelector properly.

tests/cnf/ran/ztp/tests/ztp-spoke-checker.go
klaskosk: @yliu127 I originally changed it to use all nodes but it seemed to fail running the command on the worker node and it looks like the worker node does not have the services disabled. The cnf-gotests version just checks the first node which seems to always be the control-plane

yliu127 (approved):

tests/cnf/ran/ztp/tests/ztp-spoke-checker.go
```diff
@@ -20,26 +20,31 @@ var _ = Describe("ZTP Spoke Checker Tests", Label(tsparams.LabelSpokeCheckerTest
 			By("verifying chronyd is inactive")
 			// Use `| cat -` to explicitly ignore the return code since it will be nonzero when the service
 			// is not active, which is expected here.
-			status, err := cluster.ExecCommandOnSNO(
-				Spoke1APIClient, 3, "systemctl is-active chronyd.service | cat -")
+			statuses, err := cluster.ExecCmdWithStdout(Spoke1APIClient, 3, "systemctl is-active chronyd.service | cat -")
 			Expect(err).ToNot(HaveOccurred(), "Failed to check if chronyd service is active")
+			Expect(statuses).ToNot(BeEmpty(), "Failed to find statuses for chronyd service")
```
mpmaruthu, line 25:
```suggestion
			Expect(statuses).ToNot(BeEmpty(), "Failed to find status for chronyd service")
```

tests/cnf/ran/ztp/tests/ztp-spoke-checker.go
```diff
@@ -20,26 +20,31 @@ var _ = Describe("ZTP Spoke Checker Tests", Label(tsparams.LabelSpokeCheckerTest
...
+			}
 
 			By("verifying chronyd is disabled")
 			// Use `| cat -` to explicitly ignore the return code since it will be nonzero when the service
 			// is disabled, which is expected here.
-			status, err = cluster.ExecCommandOnSNO(
-				Spoke1APIClient, 3, "systemctl is-enabled chronyd.service | cat -")
+			statuses, err = cluster.ExecCmdWithStdout(Spoke1APIClient, 3, "systemctl is-enabled chronyd.service | cat -")
 			Expect(err).ToNot(HaveOccurred(), "Failed to check if chronyd service is enabled")
+			Expect(statuses).ToNot(BeEmpty(), "Failed to find statuses for chronyd service")
```
mpmaruthu, line 39:
```suggestion
			Expect(statuses).ToNot(BeEmpty(), "Failed to find status for chronyd service")
```

mpmaruthu (approved):
2 minor nit comments otherwise looks good to me.

PR #160, opened 2024-08-14
cnf ran: reorganize powermanagement helpers

klaskosk:
This moves the powermanagement stats file to be a package under ran and adds unit tests for it. Additionally, the metric collections helpers have been moved to their own package under powermanagement.


yliu127 (approved):

tests/cnf/ran/powermanagement/internal/collect/collect.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(stressNgPods) < 1 {
+		return nil, fmt.Errorf("not enough stress-ng pods to run test")
+	}
+
+	glog.V(tsparams.LogLevel).Infof("Wait for %s for steadyworkload scenario", duration.String())
+	result, collectErr := collectPowerUsageMetrics(duration, interval, "steadyworkload", tag)
```
sebrandon1, line 76: Should `collectErr` be checked immediately after this?

sebrandon1 (approved):
A comment about an error check.

tests/cnf/ran/powermanagement/internal/collect/collect.go
klaskosk: I think this came up when the test got merged the first time lol but my idea is that we want to tear down the stress pods regardless of whether the metrics collection failed. Might be better to delete the pods in a `defer` but the preference seems to be against `defer` in tests

sebrandon1: I think `defer` is fine in tests just outside of a loop.

Sounds good though!

PR #161, opened 2024-08-14
global: Fail ExecCmdWithStdout if no nodes were found to execute command

edcdavid:
Otherwise retry functions using this will not retry if no nodes are found(transient conditon after reboot).


edcdavid:
Removing after discussion with @klaskosk

PR #162, opened 2024-08-14
global: fail fast on retry errors

klaskosk:
As originally used by the cnf/ran directory, retry is only supposed to retry on errors that come from the command execution failing, often from the MCO daemon restarting mid-execution. The current implementation will ignore errors completely that are not to be retried. This masks any errors in the actual command failing.

This PR just changes back to the original behavior and will fail fast when the command gets executed but fails, for example, by having a nonzero exit code.

### DIFF
--- a/tests/internal/cluster/cluster.go
+++ b/tests/internal/cluster/cluster.go
@@ -193,13 +193,16 @@ func ExecCmdWithRetries(client *clients.Settings, retries uint,
 	return wait.PollUntilContextTimeout(
 		context.TODO(), interval, time.Duration(retries-1)*interval, true, func(ctx context.Context) (bool, error) {
 			err := ExecCmd(client, nodeSelector, command)
-
 			if isErrorExecuting(err) {
-				glog.V(90).Infof("Error during command execution, retry %d (%d max): %w", retry, retries, err)
+				glog.V(90).Infof("Error during command execution, retry %d (%d max): %v", retry, retries, err)
 
 				retry++
 
 				return false, nil
+			}
+
+			if err != nil {
+				return false, err
 			}
 
 			return true, nil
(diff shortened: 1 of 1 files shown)


edcdavid (approved):
lgtm, matches previous function now: 
- retries if error is "error executing command in container" or "container not found"
- fail if any other error 
- completes successfully if no error

kononovn (approved):
/lgtm

tests/internal/cluster/cluster.go
```diff
@@ -193,13 +193,14 @@ func ExecCmdWithRetries(client *clients.Settings, retries uint,
...
 			err := ExecCmd(client, nodeSelector, command)
-
 			if isErrorExecuting(err) {
-				glog.V(90).Infof("Error during command execution, retry %d (%d max): %w", retry, retries, err)
+				glog.V(90).Infof("Error during command execution, retry %d (%d max): %v", retry, retries, err)
 
 				retry++
 
 				return false, nil
+			} else if err != nil {
```
trewest, line 202: Why else if ?

tests/internal/cluster/cluster.go
```diff
@@ -215,25 +216,29 @@ func ExecCmdWithStdoutWithRetries(
...
-
 			outputs, err = ExecCmdWithStdout(client, command, options...)
 			if isErrorExecuting(err) {
-				glog.V(90).Infof("Error during command execution, retry %d (%d max): %w", retry, retries, err)
+				glog.V(90).Infof("Error during command execution, retry %d (%d max): %v", retry, retries, err)
 
 				retry++
 
 				return false, nil
+			} else if err != nil {
```
trewest, line 234: same here

tests/internal/cluster/cluster.go
klaskosk: Oh this can just be a separate if since the first block always returns, I'll update it

trewest (approved):
/lgtm

PR #185, opened 2024-09-03
global: remove pointer to containerSpec

sebrandon1:
Depends-on: https://github.com/openshift-kni/eco-goinfra/pull/638

Modifies the calls to the `deployment` pkg in the eco-goinfra library to be more consistent with the `daemonset` pkg.

Bonus: I added the `depends-on` github action to the `makefile.yml` to allow for bringing in the dependent pull request as a requirement.


kononovn (approved):
/lgtm

PR #194, opened 2024-09-10
global: added all schemes to reporter package

klaskosk:
Updated the reporter package to add all of the schemes currently used in reports for this repo. Also removed the option to provide a scheme attacher to the reporter functions since it is now done by default.


trewest (approved):
/lgtm

PR #203, opened 2024-09-16
assisted ztp: update default registries in unauthenticated registries…

trewest:
… test

Updates default registries used in unauthenticated registries test. Needed due to this change -  https://github.com/openshift/assisted-service/pull/6553


vkolodny:
what is CSV registry

ddmitrie (approved):
/lgtm

PR #209, opened 2024-09-18
cnf core: use GetGVR() instead of GetServiceGVR()

sebrandon1:
Builds on: https://github.com/openshift-kni/eco-goinfra/pull/667

Bumps the version of eco-goinfra.  `GetServiceGVR()` has been deprecated in favor of a more generic `GetGVR()` to match other functions in eco-goinfra.


klaskosk (approved):
/lgtm
cc: @gkopels @evgenLevin

PR #253, opened 2024-10-11
cnf ran: clusterinstance skip aftereach when not needed

klaskosk:
Found out that this can fail in CI sometimes when the test case isn't applicable anyway. This PR just allows the AfterEach to be skipped when these tests do not apply, preventing the errors from appearing in CI. It does this by only running the AfterEach once changes are made to need it.


mpmaruthu:
Looks good to me.

mpmaruthu (approved):

dgonyier (approved):
lgtm

PR #261, opened 2024-10-18
rhwa: DAST test on FAR operator

frajamomo:
Adds a test that runs rapidast scan for the FAR operator installed in the cluster.

It configures the rapidast for the execution of the trivy scan, looking for misconfigurations on the pods running on the namespace where the operator is installed.

The test fails if any High severity vulnerability is found, providing in the console output the details of the detection.

### DIFF
--- a/tests/rhwa/internal/dast/dast.go
+++ b/tests/rhwa/internal/dast/dast.go
@@ -1,0 +1,53 @@
+package dast
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	. "github.com/onsi/gomega"
+	"github.com/openshift-kni/eco-gotests/tests/rhwa/internal/rhwaparams"
+	"gopkg.in/yaml.v3"
+)
+
+func prepareRapidastConfig(destinationPath string) error {
+
+	workingDir, err := os.Getwd()
+	Expect(err).ToNot(HaveOccurred(), err)
+
+	templateDir := filepath.Join(workingDir, rhwaparams.RapidastTemplateFolder)
+	rapidastConfigFile := filepath.Join(templateDir, rhwaparams.RapidastTemplateFile)
+
+	f, err := os.Open(rapidastConfigFile)
+	Expect(err).NotTo((HaveOccurred()))
+	defer f.Close()
+
+	content, err := io.ReadAll(f)
+	Expect(err).NotTo((HaveOccurred()))
+
+	yamlData := make(map[string]interface{})
+	err = yaml.Unmarshal(content, &yamlData)
+	Expect(err).NotTo((HaveOccurred()))
+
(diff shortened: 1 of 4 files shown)


frajamomo:
//cc @clobrano

frajamomo:
//cc @siserafin @prabinovRedhat @shellymiron @sarroutbi @anna-savina

tests/rhwa/internal/dast/dast.go
```diff
@@ -0,0 +1,83 @@
...
+	if sc == nil || !successful {
+		return fmt.Errorf("yamlData is missing key \"scanners\"")
+	}
+
+	genericTrivy, successful := sc["generic_trivy"].(map[string]interface{})
+	if genericTrivy == nil || !successful {
+		return fmt.Errorf("yamlData is missing key \"generic_trivy\"")
+	}
+
+	genericTrivy["inline"] = fmt.Sprintf("trivy k8s --kubeconfig=/home/rapidast/.kube/config "+
```
clobrano, line 51: Is this `kubeconfig` the same defined in `KUBECONFIG` env variable read in `far.go:79`? Or, should this be the same value?

tests/rhwa/internal/dast/dast.go
```diff
@@ -0,0 +1,83 @@
...
+		return err
+	}
+	defer rapidastConfig.Close()
+
+	_, err = rapidastConfig.Write(modifiedYaml)
+	if err != nil {
+		return err
+	}
+
+	err = rapidastConfig.Close()
```
clobrano, line 77: You don't need to manually close the file, as you already deferred the operation at line 70

clobrano (changes requested):

tests/rhwa/internal/dast/dast.go
frajamomo: Although both are pointing to the same kubeconfig file, this is the location for the `kubeconfig` inside the container, where the rapidast software expects it. In `far.go` it is referenced as a mounted volume for the podman command.

tests/rhwa/internal/dast/dast.go
frajamomo: Fixed

tests/rhwa/internal/dast/dast.go
clobrano: since you have the path it, I would consider passing it as argument then

clobrano: ah, nevermind. I saw now that you mount the location of the kubeconfig in podman

clobrano (approved):

tests/rhwa/internal/dast/dast.go
frajamomo: Exactly, and that location in the podman command is meant to be hardcoded, not parametrized.

PR #267, opened 2024-11-04
cnf-network: removed timers configuration from the frrconfig CR

evgenLevin:
According to the latest changes, timers should only be modified in the BgpPeer CR, not in frrconfig.


gkopels:
lgtm

PR #280, opened 2024-11-14
hwaccel: adding more log and looking for the right csv

ggordaniRed:
Fix a small issue when there more then one csvs  usually happens in Prow


cdvultur (approved):
/lgtm

wabouhamad (approved):
/lgtm

PR #330, opened 2024-12-11
separate repo folder from home folder, to allow repo override

AsherShoshan:


achuzhoy (approved):

Dockerfile
```diff
@@ -14,11 +14,16 @@ RUN dnf install -y tar gcc make && \
...
+    useradd -U -u 1000 -m -d /home/${CONTAINERUSER} -s /usr/bin/bash ${CONTAINERUSER} && \
+    mkdir /eco-gotests && chown ${CONTAINERUSER}:${CONTAINERUSER} /eco-gotests
 
 USER ${CONTAINERUSER}
 WORKDIR /home/${CONTAINERUSER}
 RUN go install github.com/onsi/ginkgo/v2/${GINKGO_VER}
+ENV GOPATH="/home/${CONTAINERUSER}/go"
+ENV PATH="$PATH:$GOPATH/bin"
+
+WORKDIR /eco-gotests
```
trewest, line 26: Instead of /eco-gotests can we put the path under the test user directory?
```suggestion
WORKDIR /home/${CONTAINERUSER}/eco-gotests
```

AsherShoshan: This will impact the OVRD (in ocp-eco-gotests.groovy) - ${CONTAINERUSER} is unknown there
/ is better.

Dockerfile
```diff
@@ -9,16 +9,21 @@ LABEL go.version=${GO_VER}
 LABEL ginkgo.version=${GINKGO_VER}
 LABEL container.user=${CONTAINERUSER}
 
-ENV PATH "$PATH:/usr/local/go/bin:/root/go/bin"
+ENV GOPATH="/usr/local/go"
+ENV PATH "$PATH:$GOPATH:$GOPATH/bin"
```
achuzhoy, line 13: do we anticipate having executables under $GOPATH 
also probably better to have $GOPATH/bin before $PATH, i.e.:
ENV PATH "$GOPATH/bin:$PATH"

Dockerfile
trewest: We maintain both so it's safe to assume the user. If it isn't dynamic enough you can use this:
```
$ skopeo inspect docker://quay.io/ocp-edge-qe/eco-gotests:latest | jq -r '.Labels["container.user"]'
testuser
```

Dockerfile
```diff
@@ -9,16 +9,21 @@ LABEL go.version=${GO_VER}
...
+
 RUN dnf install -y tar gcc make && \
     dnf clean metadata packages && \
     arch=$(arch | sed s/aarch64/arm64/ | sed s/x86_64/amd64/) && \
     curl -Ls https://go.dev/dl/${GO_VER}.linux-${arch}.tar.gz |tar -C /usr/local -xzf -  && \
-    useradd -U -u 1000 -m -d /home/${CONTAINERUSER} -s /usr/bin/bash ${CONTAINERUSER}
+    go install github.com/onsi/ginkgo/v2/${GINKGO_VER} && \
+    useradd -U -u 1000 -m -d /home/${CONTAINERUSER} -s /usr/bin/bash ${CONTAINERUSER} && \
+    mkdir /eco-gotests && chown ${CONTAINERUSER}:${CONTAINERUSER} /eco-gotests && \
+    echo 'RUN done'
```
trewest, line 22: This isn't necessary

trewest:
> Also changed the ginkgo install to be to local/usr/go, and nothing will be left in /home/testuser

@AsherShoshan we would like to contain everything to the testusers home directoy. Please add the eco-gotests folder to /home/testuser/eco-gotests

Dockerfile
AsherShoshan: What is not necessary?

Dockerfile
AsherShoshan: do u mean that inside container workdir would be $HOME/eco-gotests ?

AsherShoshan: podman inspect $ECO_GOTESTS_IMAGE | jq -r '.[].Config.WorkingDir'

Dockerfile
trewest:
```suggestion
```

AsherShoshan: Why do u think I put it?    It's a way to add commands to the RUN, and always use the && \, and worry about remove it from last line..

AsherShoshan: Please merge this.  the ocp-eco-gotests.groovy is pending on this.

AsherShoshan:
.

PR #352, opened 2025-01-07
system-test: retry commands on the nodes

yprokule:
Added extra validation and retries to the commands that are executed on the remote nodes to prevent issues with malformed output that makes tests panic:
```code
I0106 11:51:59.489310   35463 pod.go:503] Execute command [chroot /rootfs /bin/sh -c readlink /proc/$(pgrep kubelet)/ns/mnt] in the pod machine-config-daemon-qtkvl container machine-config-daemon in namespace openshift-machine-config-operator
  [PANICKED] Test Panicked
  In [It] at: /usr/local/go/src/runtime/panic.go:115 @ 01/06/25 11:51:59.661

  runtime error: index out of range [1] with length 1

  Full Stack Trace
    github.com/openshift-kni/eco-gotests/tests/system-tests/rdscore/internal/rdscorecommon.mountNamespaceEncapsulation({0xc00060bc70, 0x47})
        /home/testuser/tests/system-tests/rdscore/internal/rdscorecommon/mount_namespace.go:70 +0x1612
    github.com/openshift-kni/eco-gotests/tests/system-tests/rdscore/internal/rdscorecommon.VerifyMountNamespaceOnCNFMCP({0x5fd66e0?, 0x6174654d2065726f?})
        /home/testuser/tests/system-tests/rdscore/internal/rdscorecommon/mount_namespace.go:121 +0x28
```

### DIFF
--- a/tests/system-tests/rdscore/internal/rdscorecommon/mount_namespace.go
+++ b/tests/system-tests/rdscore/internal/rdscorecommon/mount_namespace.go
@@ -1,6 +1,7 @@
 package rdscorecommon
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"time"
@@ -10,6 +11,7 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/openshift-kni/eco-goinfra/pkg/nodes"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/openshift-kni/eco-gotests/tests/system-tests/internal/remote"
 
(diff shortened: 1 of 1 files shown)


yprokule:
> it will fail anyway, but not after the the first, but after the third time.

the issue doesn't reproduce after a retry which makes me think there's some kind of race condition

klaskosk:
Hi @yprokule @elenagerman, I've been hitting a similar issue in cnf/ran, from what I can best tell it's a case of the SPDY command executor being unstable. Kubernetes 1.30 introduced the websocket command executor which seems to fix the issue. Adding a `; sleep 0.1` to the end is another workaround. See kubernetes/kubernetes#124571

PR #369, opened 2025-01-19
cnf ran: fix talm backup automation issue

klaskosk:
Some commands executed by talm backup may not return the output occasionally, so this PR changes the behavior to retry until an output is found.

### DIFF
--- a/tests/cnf/ran/talm/internal/mount/mount.go
+++ b/tests/cnf/ran/talm/internal/mount/mount.go
@@ -29,13 +29,11 @@ func PrepareEnvWithSmallMountPoint(client *clients.Settings) (string, error) {
 	}
 
 	// find the next available loopback device (OS takes care of creating a new one if needed)
-	loopbackDevicePath, err := cluster.ExecCommandOnSNOWithRetries(client, ranparam.RetryCount, ranparam.RetryInterval,
-		"sudo losetup -f")
+	loopbackDevicePath, err := execWhileBlank(client, "sudo losetup -f")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


klaskosk:
Tested: https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/2230

tests/cnf/ran/talm/internal/mount/mount.go
```diff
@@ -188,3 +177,22 @@ func unmountLoopback(client *clients.Settings, loopbackDevicePath, findmntOutput
...
 }
+
+// execWhileBlank runs the provided command on the cluster until the output, after trimming whitespace, is not empty.
+func execWhileBlank(client *clients.Settings, command string) (string, error) {
+	var (
+		output string
+		err    error
+	)
+
+	for output == "" {
```
yliu127, line 188: I want to be caution about this - is there a chance this output would never be non-empty? or what if we keep encountering error - is output going to be empty in this case?

klaskosk: @yliu127 I debated whether to add a check to avoid that but it seems to have never been an issue with the device type command. I'll add one though since the off chance of an infinite loop would be bad

klaskosk:
Revisions tested: https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/2244

yliu127 (approved):

josclark42:
LGTM

PR #370, opened 2025-01-20
infra: add systemreporter pkg

trewest:
- Adds systemreporter pkg
- Adds ability to gather command output from machine under test through kube-api 
- Adds ability to gather command output from machine under test through ssh connection
- Adds ECO_SSH_KEY_PATH and ECO_SSH_USER to GeneralConfig in order to to provide SSH access to nodes under test

### DIFF
--- a/tests/internal/systemreporter/systemreporter.go
+++ b/tests/internal/systemreporter/systemreporter.go
@@ -1,0 +1,187 @@
+package systemreporter
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	re "regexp"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/onsi/ginkgo/v2/types"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-gotests/tests/internal/cluster"
+	. "github.com/openshift-kni/eco-gotests/tests/internal/inittools"
+	"golang.org/x/crypto/ssh"
+)
+
+var (
+	// Matches option hypens, spaces, and special characters.
+	specialChars = re.MustCompile(`-?\s-?|[/|'"\.\[\]]`)
+
+	// Matches duplicate underscores.
+	dupUnderscores = re.MustCompile(`__+`)
+
+	// Matches leading and trailing underscores.
+	leadAndTrailUnderscores = re.MustCompile(`^_|_$`)
+)
+
+// ReportIfFailedFromClient dumps the requested command output
+// from nodes pulled from specified apiClient if test case fails.
+func ReportIfFailedFromClient(
+	report types.SpecReport, testSuite string, commands []string, apiClient *clients.Settings) {
+	if types.SpecStateFailureStates.Is(report.State) {
+		if apiClient == nil {
+			glog.Errorf("cannot gather system report from nil apiClient")
+
+			return
+		}
+
+		dumpDir := GeneralConfig.GetDumpFailedTestReportLocation(testSuite)
+
+		tcReportFolderName := strings.ReplaceAll(report.FullText(), " ", "_")
+
+		systemFolder := fmt.Sprintf("%s/%s/system", dumpDir, tcReportFolderName)
+
+		err := os.MkdirAll(systemFolder, 0755)
+		if err != nil {
+			glog.Errorf("failed creating dir for system info: %s", err)
+
+			return
+		}
+
+		GatherInfoThroughKubeClient(commands, systemFolder, apiClient)
+	}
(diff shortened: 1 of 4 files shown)


tests/internal/systemreporter/systemreporter.go
```diff
@@ -0,0 +1,187 @@
...
+			return
+		}
+
+		dumpDir := GeneralConfig.GetDumpFailedTestReportLocation(testSuite)
+
+		tcReportFolderName := strings.ReplaceAll(report.FullText(), " ", "_")
+
+		systemFolder := fmt.Sprintf("%s/%s/system", dumpDir, tcReportFolderName)
+
+		if _, err := os.Stat(systemFolder); os.IsNotExist(err) {
```
klaskosk, line 35: `os.MkdirAll` is a no-op if the path already exists so this check is unnecessary

tests/internal/systemreporter/systemreporter.go
```diff
@@ -0,0 +1,187 @@
...
+// ReportIfFailedFromNodeList dumps the requested command output from specified nodes through SSH if test case fails.
+func ReportIfFailedFromNodeList(report types.SpecReport, testSuite string, commands []string, nodes []string) {
+	if types.SpecStateFailureStates.Is(report.State) {
+		dumpDir := GeneralConfig.GetDumpFailedTestReportLocation(testSuite)
+
+		tcReportFolderName := strings.ReplaceAll(report.FullText(), " ", "_")
+
+		systemFolder := fmt.Sprintf("%s/%s/system", dumpDir, tcReportFolderName)
+
+		if _, err := os.Stat(systemFolder); os.IsNotExist(err) {
```
klaskosk, line 57: see above, this check is unnecessary

tests/internal/systemreporter/systemreporter.go
```diff
@@ -0,0 +1,187 @@
...
+	fileName = dupUnderscores.ReplaceAll(fileName, []byte("_"))
+
+	// Remove leading and trailing underscores
+	leadAndTrailUnderscores := re.MustCompile(`^_|_$`)
+	fileName = leadAndTrailUnderscores.ReplaceAll(fileName, []byte(""))
+
+	return string(fileName)
+}
+
+func writeFile(content []byte, file string) error {
```
klaskosk, line 185: this function seems a little unnecessary

tests/internal/systemreporter/systemreporter.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+}
+
+func fileNameFromCommand(command string) string {
+	fileName := []byte(command)
+
+	// Replace option hypens, spaces, and special characters with underscores everywhere in the command.
+	specialChars := re.MustCompile(`-?\s-?|[/|'"\.\[\]]`)
+	fileName = specialChars.ReplaceAll(fileName, []byte("_"))
```
klaskosk, line 172: since all the inputs start as string and this function returns a string, `ReplaceAllString` will prevent so many conversions to and from `[]byte`

tests/internal/systemreporter/systemreporter.go
```diff
@@ -0,0 +1,187 @@
...
+			}
+		}
+	}
+}
+
+func fileNameFromCommand(command string) string {
+	fileName := []byte(command)
+
+	// Replace option hypens, spaces, and special characters with underscores everywhere in the command.
+	specialChars := re.MustCompile(`-?\s-?|[/|'"\.\[\]]`)
```
klaskosk, line 171: nit since this isn't performance sensitive, but by compiling in this function the regex is compiled for every node for every command whereas storing the regexp as a variable outside the function prevents duplicating the compile

klaskosk (approved):
some minor comments, overall lgtm

tests/internal/systemreporter/systemreporter.go
trewest: Thanks - addressed

tests/internal/systemreporter/systemreporter.go
trewest: Thanks - addressed

tests/internal/systemreporter/systemreporter.go
trewest: Thanks - addressed

tests/internal/systemreporter/systemreporter.go
trewest: Thanks - addressed

achuzhoy (approved):

klaskosk (approved):
/lgtm

PR #373, opened 2025-01-21
lca ibu: add the ability to check is the cluster was deployed with fips

achuzhoy:
This requires moving the installconfig package upper in the tree.


achuzhoy:
cc @trewest 
Successfully tested.

trewest (approved):
/lgtm

PR #381, opened 2025-01-23
cnf ran: fix ZTP suite automation issues

klaskosk:
* Add a workaround for an issue where the policy definition would be nil. Failures only commonly occur during this ocm.PullPolicy specifically so a workaround here is sufficient.
* Switches the custom source CR service account namespace to the test namespace which requires creating the namespace on the spoke too.
* Change to recreating the test namespace per test case rather than per suite. This should improve the isolation between tests and fix some of the automation issues related to namespaces.


klaskosk:
Tested: https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/2268

yliu127 (approved):

PR #383, opened 2025-01-27
save original boot order instead of reading current value and swapping when recovering

edcdavid:
In latest tests (new dell platform), reading swapped boot order seems to return wrong value (not swapped) on second boot but looks correct in the bios. So if just reading and swapping back the boot order with lead to no changes (e.g. not swapped -> swapped again).

Instead of reading latest boot order and swapping back and forth, back up original boot order and restore it after testing with swapped boot order.

### DIFF
--- a/tests/system-tests/diskencryption/tests/tpm2.go
+++ b/tests/system-tests/diskencryption/tests/tpm2.go
@@ -5,6 +5,7 @@ import (
 
 	"regexp"
 
+	"github.com/golang/glog"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/openshift-kni/eco-goinfra/pkg/nodes"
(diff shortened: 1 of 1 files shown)


tests/system-tests/diskencryption/tests/tpm2.go
```diff
@@ -413,6 +413,22 @@ var _ = Describe("TPM2", func() {
...
+			if len(originalBootOrder) < 2 {
+				return fmt.Errorf("need at least 2 boot entries")
+			}
+
+			return err
+		}).WithTimeout(tsparams.TimeoutWaitingOnBMC).
+			WithPolling(tsparams.PollingIntervalBMC).
+			ShouldNot(HaveOccurred(), "getting boot order should not return an error")
+
+		By(fmt.Sprintf("listing original boot Order: %s", originalBootOrder))
```
klaskosk, line 430: this should be a log instead of By

klaskosk (commented):
one minor comment but lgtm assuming it's been tested

tests/system-tests/diskencryption/tests/tpm2.go
edcdavid: Fixed

klaskosk (approved):
/lgtm

agopala95 (commented):
Assuming the original boot order holds up during reboots and reconfigurations (could be a silly thought), lgtm!

agopala95 (approved):

edcdavid:
> Assuming the original boot order holds up during reboots and reconfigurations (could be a silly thought), lgtm!

cool! Just this limited swapping of the boot entry with no other changes should be pretty safe here

PR #393, opened 2025-01-30
reporter: add oran apis to schemes

klaskosk:
* Adds the schemes for ORAN CRs to the reporter package. These are spread across multiple packages that correspond to the ones used by the ORAN eco-goinfra PRs.
* Vendors the ORAN apis and their dependencies. These will already be included with the eco-goinfra resources once it is bumped.


trewest (approved):
/lgtm

PR #394, opened 2025-01-30
System Test, RDS Core: active PDB failure on drain work-around

elenagerman:


elenagerman:
@yprokule PTAL, this is eIP failover active PDB on Drain failure workaround

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
```diff
@@ -496,7 +499,7 @@ func gracefulNodeReboot(nodeName string) error {
 		return fmt.Errorf("failed to retrieve node %s object due to: %w", nodeName, err)
 	}
 
-	glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Cordoning node %q", nodeName)
+	glog.V(100).Infof("Cordoning node %q", nodeName)
```
yprokule, line 502: any particular reason to switch from `rdscoreparams.RDSCoreLogLevel` variable?

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
```diff
@@ -672,6 +681,40 @@ func getNodeForReboot(isIPv6 bool) (string, string, error) {
...
 
+func setMinAvailableToZeroForActivePDB() error {
+	By("Workaround for the node drain failure due to active PodDisruptionBudget")
+
+	allAvailablePDB, err := poddisruptionbudget.ListInAllNamespaces(APIClient)
+
+	if err != nil {
+		glog.Infof("Failed to list all available pod disruption budget due to %v", err)
+
+		return nil
```
yprokule, line 692: if there's an error listing PDBs in all the namespaces, why it's not returned outside?

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
```diff
@@ -672,6 +681,40 @@ func getNodeForReboot(isIPv6 bool) (string, string, error) {
...
+	}
+
+	zeroInt := intstr.FromInt32(0)
```
yprokule, line 695: why not to use regular type conversion:
```
zeroInt := int32(0)
```

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
```diff
@@ -672,6 +681,40 @@ func getNodeForReboot(isIPv6 bool) (string, string, error) {
 	return "", "", fmt.Errorf("no egress IP address found in egressIP map")
 }
 
+func setMinAvailableToZeroForActivePDB() error {
```
yprokule, line 684: All changed PDBs should be restored to respective  initial value after the tests

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
elenagerman: yes, I always use the general default value (100) for sub-functions. it will be easy to handle it if will be decided to move some methods to the upper level

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
elenagerman: because if there is no pdb found we don't care

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
elenagerman: because it is not suitable for the required variable type:Cannot use '&zeroInt' (type *int32) as the type *intstr. IntOrString

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
elenagerman: it is restored automatically by cluster. this is what I meant when told you that in this case it is not required.
but can be required in case of update/upgrade

yprokule: > it is restored automatically by cluster. this is what I meant when told you that in this case it is not required.
but can be required in case of update/upgrade

I'm a bit confused by this statement - how could cluster reset PDB configuration when it's modified by user? Is this expected behavior? does it affect all PDBs or only some?

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/nmstate/kubernetes-nmstate/api v0.0.0-20250114063637-129b149f6ce9
 	github.com/onsi/ginkgo/v2 v2.22.2
 	github.com/onsi/gomega v1.36.2
-	github.com/openshift-kni/eco-goinfra v0.0.0-20250129204500-1fb7f3c38dc7 // latest
+	github.com/openshift-kni/eco-goinfra v0.0.0-20250130220349-f088d0044ca1 // latest
```
yprokule, line 25: shouldn't this be done in a dedicated PR?

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
elenagerman: Our problem during worker node drain is openshift-monitoring pdb. I cannot provide you too much details, but probably there is some kind of mechanism for the default pdb values restoring. it take min or so:

```
[kni@registry ~]$ oc get pdb metrics-server -n openshift-monitoring
NAMESPACE                              NAME                                    MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
openshift-monitoring                   metrics-server                          1               N/A               2                     14h
[kni@registry ~]$ 
[kni@registry ~]$ 
[kni@registry ~]$ oc edit pdb metrics-server -n openshift-monitoring
poddisruptionbudget.policy/metrics-server edited
[kni@registry ~]$ 
[kni@registry ~]$ 
[kni@registry ~]$ oc get pdb -A
NAMESPACE                              NAME                                    MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
openshift-monitoring                   metrics-server                          0               N/A               2                     14h
[kni@registry ~]$ 
[kni@registry ~]$ 
[kni@registry ~]$ oc get pdb metrics-server -n openshift-monitoring
NAMESPACE                              NAME                                    MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
openshift-monitoring                   metrics-server                          1               N/A               2                     14h
[kni@registry ~]$ 

```

go.mod
elenagerman: I don't think so, I do it inside same mr all the time. it has different commit message : "bump vendors"
this is how Nikita explained it too me long time ago

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
elenagerman: if you think that we need to add step to restore something that does not require restore I have no any problem to add it. I have this method implemented once, it just was removed as unnecessary

tests/system-tests/rdscore/internal/rdscorecommon/egress-ip.go
```diff
@@ -626,23 +635,31 @@ func gracefulNodeReboot(nodeName string) error {
...
 	if err != nil {
-		glog.V(rdscoreparams.RDSCoreLogLevel).Infof("Failed to uncordon %q due to %v", nodeName, err)
+		glog.V(100).Infof("Failed to uncordon %q due to %v", nodeName, err)
 
 		return fmt.Errorf("failed to uncordon %q due to %w", nodeName, err)
 	}
 
 	time.Sleep(15 * time.Second)
 
+	err = restoreActivePDBValues(pdbList)
```
yprokule, line 655: this part won't be executed in case there's any preceding error (draining/uncording/etc) so more appropriate place for this will be in _BeforeAll_ to capture initial state of PDBs and _Before/AfterEach_ to restore them, wdyt?

elenagerman: I think it's a great idea, but I doubt how I can implement it. The problem is pdbList, all info for this List was collected during set... state

elenagerman: actually I can separate 'collect pdb info' into stand alone function, wait a min, I'll check...

elenagerman: no, it will not help. I still do not see how I can pass to the AfterEach state PDB list. Do you have any idea?

elenagerman: @yprokule do you have any idea how to implement it?

yprokule: Sent U some draft directly. Also noticed that some _PDBs_ have their _minAvailable_ field reverted after it's updated by the test, this must be investigated further.

Also noticed that after node reboot it takes some time for respective pods to start so proposed an extra check here - https://github.com/openshift-kni/eco-gotests/pull/413

ghost:
@manurodriguez 
ugly

How dare you lay your dirty hands on newborn babies while having an affair with a married woman/your colleague on the same team?

Squatting the apartment with the woman you're having affair?

Having affairs under the disguise of attending a conference?

Your appearance truly reflects your needs, disgusting and revolting.

You even have the courage to show your dick face anywhere? In the office, in front of your colleagues?

Did you pray to your God and ask for forgiveness?

A rat has more value in life than you do.

You should never have been born into this world. The animals that reproduced you should have left you in a toilet 44 years ago.

Do you think becoming a Canadian would whitewash your sins?

Your true colors cannot be hidden. Your disgusting appearance, your lousy accent, and, above all, your lack of morality reveal who you truly are.

For the rest of your numbered days, may your God show you some mercy, so that you can sleep peacefully at night or look people in the eyes, hoping that they do not know about your disgusting deeds.

PR #413, opened 2025-02-05
system-test: wait for pod to be ready for EgressIP test

yprokule:
After node is rebooted it takes some time for the pod to start hence extra check is added to ensure pod is `Ready` before trying to execute commands from it.


elenagerman:
we definitely can add such a verification, but it will make sense only in case we'll have some work-around for the bdp issue

PR #438, opened 2025-02-13
system-test: increase timeout while executing command

yprokule:
Invoked command has a timeout of 5 seconds, `--connect-timeout 5`, therefor extending timeout of the retry block to 1 minute


elenagerman:
sure, it make sense!
/lgtm

PR #440, opened 2025-02-13
cnf ran: oran provisioning test cases

klaskosk:


tests/cnf/ran/oran/tests/oran-provision.go
```diff
@@ -1 +1,134 @@
...
+		bmcSecret, err := secret.Pull(HubAPIClient, clusterInstanceNode.BmcCredentialsName.Name, RANConfig.Spoke1Name)
+		Expect(err).ToNot(HaveOccurred(), "Failed to pull spoke 1 BMC secret")
+
+		bmcUsername, exists := bmcSecret.Definition.Data["username"]
+		Expect(exists).To(BeTrue(), "Username does not exist in BMC secret")
+
+		bmcUsername = bytes.TrimSpace(bmcUsername)
+		Expect(string(bmcUsername)).To(Equal(RANConfig.BMCUsername), "ClusterInstance has incorrect BMC username")
+
+		bmcPassword, exists := bmcSecret.Definition.Data["password"]
```
dgonyier, line 128: I think the bmc checks should be moved earlier. Also I think policies are applied last, so it would be better to check things in order of how the node is provisioned, i.e. siteconfig things like secrets and bmh are applied before policies are applied.

Unless the order matters to the test logic in some way.

tests/cnf/ran/oran/tests/oran-provision.go
```diff
@@ -1 +1,134 @@
...
+
+		By("saving the spoke 1 admin kubeconfig")
+		kubeconfigSecret, err := secret.Pull(HubAPIClient, RANConfig.Spoke1Name+"-admin-kubeconfig", RANConfig.Spoke1Name)
+		Expect(err).ToNot(HaveOccurred(), "Failed to get the spoke 1 kubeconfig secret")
+
+		kubeconfig, exists := kubeconfigSecret.Definition.Data["kubeconfig"]
+		Expect(exists).To(BeTrue(), "Kubeconfig key does not exist in kubeconfig secret")
+
+		err = os.WriteFile(RANConfig.Spoke1Kubeconfig, kubeconfig, 0644)
+		Expect(err).ToNot(HaveOccurred(), "Failed to save the spoke 1 admin kubeconfig")
```
yliu127, line 42: We shall save admin password as well.

tests/cnf/ran/oran/tests/oran-provision.go
```diff
@@ -1 +1,134 @@
...
+// become valid again, we cannot test provisioning with a valid ProvisioningRequest.
+var _ = Describe("ORAN Provision Tests", Label(tsparams.LabelProvision), Ordered, func() {
+	var prBuilder *oran.ProvisioningRequestBuilder
+
+	AfterAll(func() {
+		By("verifying the spoke 1 kubeconfig path is set")
+		Expect(RANConfig.Spoke1Kubeconfig).
+			ToNot(BeEmpty(), "KUBECONFIG environment variable must be set to save spoke 1 kubeconfig")
+
+		By("saving the spoke 1 admin kubeconfig")
```
yliu127, line 34: Is this only needed after successful deployment? If so, this should be in AfterEach of that specific test case.

tests/cnf/ran/oran/tests/oran-provision.go
```diff
@@ -1 +1,134 @@
...
+
+		var err error
+		prBuilder, err = prBuilder.Create()
+		Expect(err).ToNot(HaveOccurred(), "Failed to create an invalid ProvisioningRequest")
+
+		By("checking the ProvisioningRequest status for a failure")
+		prBuilder, err = prBuilder.WaitForCondition(tsparams.PRValidationFailedCondition, time.Minute)
+		Expect(err).ToNot(HaveOccurred(), "Failed to wait for the ProvisioningRequest to fail")
+
+		By("updating the ProvisioningRequest with valid policyTemplateParameters")
```
yliu127, line 63: This should be moved to the next test case, as each test case should be executable on its own even if they are under Ordered container, currently if we try to run 77394 alone, it seems the test will fail as the provisioning request has not been created yet.  

And if we do the move, then we need to check if there are existing provisioning request to decide whether we will update or create new in 77394.

tests/cnf/ran/oran/tests/oran-provision.go
klaskosk: @dgonyier The only ordering that really matters here is making sure the ProvisioningRequest is finished before running the checks since that does simplify the logic quite a bit. But after that, I see your point, the checks are pretty randomly ordered. I'll change it to what you said, I like that ordering and it's a lot more coherent than what's there now

tests/cnf/ran/oran/tests/oran-provision.go
```diff
@@ -1 +1,135 @@
...
+		By("creating a ProvisioningRequest with invalid policyTemplateParameters")
+		prBuilder = helper.NewProvisioningRequest(HubAPIClient, tsparams.TemplateValid).
+			WithTemplateParameter(tsparams.PolicyTemplateParamsKey, map[string]any{
+				// By using an integer when the schema specifies a string we can create an invalid
+				// ProvisioningRequest without being stopped by the webhook.
+				tsparams.TestName: 1,
+			})
+
+		var err error
+		prBuilder, err = prBuilder.Create()
```
yliu127, line 56: Since we rely on cleanup done by external code via CI, we probably should check if the env is clean in BeforeAll or BeforeEach for this test case, and skip if not?

tests/cnf/ran/oran/tests/oran-provision.go
```diff
@@ -1 +1,135 @@
...
+		pgMap, err := configmap.Pull(
+			HubAPIClient, RANConfig.Spoke1Name+"-pg", ztpNamespace)
+		Expect(err).ToNot(HaveOccurred(), "Failed to pull policy ConfigMap for spoke 1")
+		Expect(pgMap.Object).ToNot(BeNil(), "Failed to get policy ConfigMap object for spoke 1")
+
+		By("verifying all the policies are compliant")
+		err = ocm.WaitForAllPoliciesComplianceState(
+			HubAPIClient, policiesv1.Compliant, time.Minute, runtimeclient.ListOptions{Namespace: RANConfig.Spoke1Name})
+		Expect(err).ToNot(HaveOccurred(), "Failed to verify all spoke 1 policies are compliant")
+
```
yliu127, line 133: As we are verifying multiple things post deployment, it would be better to not fail right away, but do all the checks then fail (error should be printed after each failure though), so if one fail, the rest of the verification won't be blocked.

tests/cnf/ran/oran/tests/oran-provision.go
klaskosk: @yliu127 This block has to be part of 77393 since the test itself is verifying that we can update an invalid ProvisioningRequest and have it become valid again. I could add logic to create or update in 77394 but that would make it similar to most of 77393.

Ultimately the issue is that these two cases are like one test case split into two parts rather than independent test cases, but this is done in the automation so that we only have to do one provision. I can pull the ProvisioningRequest in 77394 and that'd make it easier to rerun just that case if it fails

yliu127: @klaskosk sounds good. We can just update 77394 to check if valid ProvisioningRequest already exists, create if not.

tests/cnf/ran/oran/internal/helper/helper.go
```diff
@@ -18,7 +18,7 @@ func NewProvisioningRequest(client *clients.Settings, templateVersion string) *o
 	versionWithAffix := RANConfig.ClusterTemplateAffix + "-" + templateVersion
 	prBuilder := oran.NewPRBuilder(client, tsparams.TestPRName, tsparams.ClusterTemplateName, versionWithAffix).
 		WithTemplateParameter("nodeClusterName", RANConfig.Spoke1Name).
-		WithTemplateParameter("oCloudSiteId", RANConfig.Spoke1Name).
+		WithTemplateParameter("oCloudSiteId", "rdu3").
```
klaskosk, line 21: after speaking with Don I learned this has to be the same as the site in the DTIAS (when using the dell adaptor), which we'll always configure as rdu3 (since there can only be one)

dgonyier (approved):

PR #445, opened 2025-02-17
report: add ginkgo report templates

klaskosk:
To break up the eco-gotests reporter into managable PRs, I have started with the templates for the report as a whole and the individual trees (branches).

report_template.html is the entrypoint for viewing the reports and simply contains a list of the branches in the report along with links to the latest commit on each branch.

tree_template.html is the report for an individual branch. It contains a tree listing out all of the specs under all of the subdirectories of tests/. Since it uses the native html details and summary tags it is entirely tab-navigable and fairly accessible.

Both templates are fairly barebones with the intention being that future PRs may add more details as desired.

---

This is PR 1/4. An example report can be seen on [my GitHub pages](https://klaskosk.github.io/eco-gotests/report.html).


trewest (approved):
/lgtm

cdvultur (approved):
/lgtm

PR #446, opened 2025-02-17
report: add most of report program

klaskosk:
Depends-on: #445

This PR contains about half of the report program as part of splitting it up into more managable PR sizes. It includes the following files:

* `command.go`: Wrapper around executing local commands such as git and ginkgo.
* `sum.go`: Small utility for getting the SHA-256 sum of the source code.
* `template.go`: Contains types and functions for executing the report templates.
* `tree.go`: Defines the SuiteTree type and methods for creating one from a Ginkgo report.

The excluded files are necessary to actually run the program, namely the cache and main files. However, these combined are quite large and depend on all of the files from this PR.

---

This is PR 2/4. An example report can be seen on [my GitHub pages](https://klaskosk.github.io/eco-gotests/report.html).


internal/report/command.go
```diff
@@ -0,0 +1,176 @@
...
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+
+	"github.com/golang/glog"
+)
+
+// CloneRepo clones the eco-gotests repo from the given repo and branch and returns the path to the cloned repo.
+func CloneRepo(ctx context.Context, localPath, repo, branch string) (string, error) {
```
trewest, line 18: Why do we need git commands rather than running directly on the project?

klaskosk: @trewest This allows you to run the program on different branches regardless of the state of the local repo. I went with a model where it always uses the program from the main branch when generating the report which keeps it up to date with the latest template and eliminates the need for backporting every change to the reporter

cdvultur (approved):
/lgtm

trewest (approved):
/lgtm

PR #447, opened 2025-02-17
report: add rest of report program

klaskosk:
Depends-on: #446

This PR contains the rest of the report program. It includes the last two files not part of the last PR:

* `cache.go`: Caches the trees generated from each branch and revision combination. Since a dry-run may take many minutes, caching allows the program to remain performant and responsive to changes in the templates without having to redo a dry-run.
* `main.go`: Handles the command line flags and actual running of the program. It handles retrieving values from the cache or cloning and dry running as necessary.

There are some small changes to the lint config and go.mod to allow the cache to be compressed. It uses a package already vendored so does not include additional dependencies. Compression is beneficial since the cache for the main branch alone may be over a megabyte in size but only ~40 KB when compressed.

Excluded from this PR is the GitHub workflow to actually generate the report along with some high-level documentation.

---

This is PR 3/4. An example report can be seen on [my GitHub pages](https://klaskosk.github.io/eco-gotests/report.html).

### DIFF
--- a/internal/report/cache.go
+++ b/internal/report/cache.go
@@ -1,0 +1,460 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"maps"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/klauspost/compress/zstd"
+)
+
+const (
+	cacheDir  = "eco-gotests"
+	remoteURL = "https://github.com/openshift-kni/eco-gotests.git"
+)
+
+var (
+	errCacheMiss = fmt.Errorf("cache miss")
+)
+
+// IsMiss returns true if the given error is a cache miss error and false otherwise.
+func IsMiss(err error) bool {
+	return errors.Is(err, errCacheMiss)
+}
+
+// CacheKey holds the information necessary to index the cached trees. Using both the branch and revision allows for
+// easily verifying if the tree cached for a given branch is current.
+type CacheKey struct {
+	Branch   string
+	Revision string
+}
+
+// Cache represents the format of the cache file. It will be saved as JSON according to the XDG base directory
+// specification.
+type Cache struct {
+	Trees     map[CacheKey]*SuiteTree
+	directory string
+	ctx       context.Context
+}
+
+// NewCacheContext creates a new cache instance. It will attempt to load the cache from the cache file. If the file does
+// not exist, a new cache will be created but not saved until Save is called.
+func NewCacheContext(ctx context.Context) (*Cache, error) {
+	glog.V(100).Info("Instantiating new Cache and attempting to load")
+
+	cache := &Cache{
+		Trees: make(map[CacheKey]*SuiteTree),
+		ctx:   ctx,
+	}
+
+	err := cache.Load()
+	if err != nil {
+		return nil, err
+	}
+
+	return cache, nil
+}
+
+// CleanCache cleans the existing cache on disk by removing the entire eco-gotests cache directory.
+func CleanCache() error {
+	cache := &Cache{ctx: context.TODO()}
+
+	cachePath, err := cache.getDirectory()
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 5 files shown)


cdvultur (approved):
/lgtm

internal/report/cache.go
```diff
@@ -0,0 +1,454 @@
...
+func NewCacheContext(ctx context.Context) (*Cache, error) {
+	cache := &Cache{
+		Trees: make(map[CacheKey]*SuiteTree),
+		ctx:   ctx,
+	}
+
+	cachePath, err := cache.getDirectory()
+	if err != nil {
+		return nil, err
+	}
```
trewest, line 58: Isn't this redundant with L93-96?

klaskosk: @trewest true this is only for the log and we log in `Load` anyway so I'll update to remove the redundancy, good catch

internal/report/cache.go
```diff
@@ -0,0 +1,454 @@
...
+		return nil, errCacheMiss
+	}
+
+	return tree, nil
+}
+
+// GetOrCreate returns the suite tree for the given repo path from the cache. It first calls Get and if there is a cache
+// miss, it calls the given create function and adds the result to the cache. Note that if the repo has local changes,
+// the create function will always be called, but the result will not be added to the cache.
+func (cache *Cache) GetOrCreate(repoPath string, create func() (*SuiteTree, error)) (*SuiteTree, error) {
```
trewest, line 235: Is the create parameter necessary? I see that there is only one function making use of this method so couldn't we just move the logic here?

internal/report/main.go
```diff
@@ -0,0 +1,308 @@
+/*
+Report is a tool to generate a report of the test suites in a Ginkgo test suite. It will print a tree of the test suites
+and the number of specs in each suite. If an output directory is provided, a static site for visualizing the test suites
+will be generated.
+
+Upon successful generation of the report the exit code is 0. If any error occurs it will be logged to stderr and the
+exit code will be 0.
```
trewest, line 7: Did you mean this?
```suggestion
exit code will be 1.
```

internal/report/cache.go
```diff
@@ -0,0 +1,449 @@
...
+		return "", err
+	}
+
+	cachePath := filepath.Join(userCache, cacheDir)
+
+	return cachePath, nil
+}
+
+// deleteExpiredFiles will delete all regular files in the cache directory that are not included in the current cache.
+// It returns early on any error deleting files.
```
trewest, line 344: So will the cache contain one file per branch?

klaskosk: yes, although my mind is blanking on why I ended up going with that (since the original was one file overall). I'll think a little more about whether this is the best approach but I can always update it later if we're wanting to merge since this does work

trewest: No worries, I was just making sure I was understanding it correctly

internal/report/cache.go
```diff
@@ -0,0 +1,460 @@
...
+
+// GetKeyFromPath returns the cache file name that corresponds to the repo at repoPath. It returns a cache miss error if
+// the repo has uncommitted changes and a different error if no source code sum is available.
+func (cache *Cache) GetKeyFromPath(repoPath string) (CacheKey, error) {
+	glog.V(100).Infof("Getting cache key for repo %s", repoPath)
+
+	changes, err := HasLocalChanges(cache.ctx, repoPath)
+	if err != nil {
+		return CacheKey{}, err
+	}
```
trewest, line 316: Why would we see uncommitted changes? If the idea of this tool is to just clone a repo/branch and compare it to what is cached, when would we run into this scenario?

klaskosk: @trewest since I started this as a cli tool, this is for running it locally against a repo with uncommitted changes. You're right that we shouldn't encounter this in our CI though

trewest (dismissed):
/lgtm

cdvultur (approved):
/lgtm

Might require a rebase, since there seems to be an conflict on `.golangci.yml`

PR #448, opened 2025-02-17
report: add workflow and documentation

klaskosk:
Depends-on: #447

The final PR in the series adding the initial version of the report program, this PR adds the GitHub workflow and documentation for the report program. Once merged, all pushes will update the report using the latest version of the program and templates.

---

This is PR 4/4. An example report can be seen on [my GitHub pages](https://klaskosk.github.io/eco-gotests/report.html).

### DIFF
--- a/.github/workflows/generate-report.yml
+++ b/.github/workflows/generate-report.yml
@@ -1,0 +1,77 @@
+name: Generate Tests Report
+
+on:
+  workflow_dispatch:
+
+  push:
+    branches:
+      - main
+      - "release-[0-9].[0-9][0-9]"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      SHELL: /bin/bash
+      XDG_CACHE_HOME: ${{ github.workspace }}/cache
+
+    steps:
+      - name: Set up Go 1.23
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.23.6
+
+      # Since we only use the checked out repo for the report generation code,
+      # always grab the latest from the default branch rather than the branch
+      # that triggered this action.
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+
+      - name: Install ginkgo
+        run: make install-ginkgo
+
+      # Actions cache is immutable, so to prevent generating a new cache every
+      # run when we only care about the latest cache, use an artifact instead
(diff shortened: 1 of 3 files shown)


.github/workflows/generate-report.yml
```diff
@@ -0,0 +1,76 @@
...
+        env:
+          ACTION_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+
+      - name: Upload generated report as artifact
+        id: upload-pages
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: ./report
+
+      - name: Save cached report data
```
sebrandon1, line 54: Would this action fit this better?  https://github.com/actions/upload-artifact

If I follow your job tasks you're trying to save the output right?

klaskosk: The report itself is saved as an artifact a couple lines above but this step is just to save the report program's cache (currently just the outputs of dry-run on each branch). Since I'm using a constant cache key anyway it may not be much different than just using the upload artifact action with `overwrite: true` though

I could use an artifact instead so if you think it fits better for this use case I can switch over to it

cdvultur (approved):
/lgtm

internal/report/README.md
```diff
@@ -0,0 +1,80 @@
...
+1. Flags are parsed.
+1. If help flag specified, help is printed and program exits.
+1. If clean flag specified, cache is cleaned and program exits.
+1. Trees are generated based on the branch flag.
+    1. If branch flag nonempty, attempt to get trees for all branches matching the patterns. Trees not present in the cache get cloned and have a dry run performed.
+    1. If branch flag empty, attempt to get trees from the repo in the current directory. Cache is checked for the current directory and a clone and dry run is performed if necessary.
+    1. Once updated, the cache is saved before any processing of the trees.
+    1. Trees are trimmed and sorted to clean them up for displaying.
+1. Trees are printed to stdout.
+1. If output flag nonempty, the generated tree map is used to fill in the templates.
```
trewest, line 67: Looks like numbering needs to be corrected

klaskosk: @trewest markdown will render the lists with sequential numbers even if you just do all 1, but this way there's no risk of it getting out of sync if we insert an item in the middle

trewest (dismissed):
/lgtm

PR #461, opened 2025-02-20
cnf ran: add oran post-provision test cases

klaskosk:


dgonyier (dismissed):

tests/cnf/ran/oran/tests/oran-post-provision.go
```diff
@@ -1 +1,315 @@
...
+		_, err := configmap.Pull(Spoke1APIClient, tsparams.TestName2, tsparams.TestName)
+		Expect(err).To(HaveOccurred(), "Second test ConfigMap already exists on spoke 1")
+
+		By("updating the ProvisioningRequest TemplateVersion")
+		prBuilder.Definition.Spec.TemplateVersion = RANConfig.ClusterTemplateAffix + "-" + tsparams.TemplateUpdateSchema
+		prBuilder = updatePRUntilNoConflict(prBuilder)
+
+		waitForPolicies(prBuilder)
+
+		By("verifying the test ConfigMap has the original value")
```
yliu127, line 175: Just try to understand the test, what did we update? Why the expected results is the same as previous add new test case? 

Perhaps it would be good to add test configs to a public repo and link somewhere to provide context.

tests/cnf/ran/oran/tests/oran-post-provision.go
```diff
@@ -1 +1,315 @@
...
+		verifyCM(tsparams.TestName, tsparams.TestOriginalValue)
+
+		By("verifying the second test ConfigMap has the new value")
+		verifyCM(tsparams.TestName2, tsparams.TestNewValue)
+	})
```
yliu127, line 180: when we revert the PR in AfterEach, does the second configmap gets deleted?

tests/cnf/ran/oran/tests/oran-post-provision.go
```diff
@@ -1 +1,315 @@
...
+
+		clusterInstanceParams, ok := templateParameters[tsparams.ClusterInstanceParamsKey].(map[string]any)
+		Expect(ok).To(BeTrue(), "Spoke 1 clusterInstanceParameters is not a map[string]any")
+
+		clusterInstanceParams["extraLabels"] = map[string]any{"ManagedCluster": map[string]string{tsparams.TestName: ""}}
+		prBuilder = prBuilder.WithTemplateParameters(templateParameters)
+
+		prBuilder = updatePRUntilNoConflict(prBuilder)
+		waitForLabels()
+	})
```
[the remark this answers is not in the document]
klaskosk: Good catch, this is an oversight on my part. I realized that removing the label from the ClusterInstance didn't remove it from the ManagedCluster until after I wrote it so it originally just asserted the label didn't exist. I'll update the AfterEach

tests/cnf/ran/oran/tests/oran-post-provision.go
```diff
@@ -1 +1,315 @@
...
+
+		By("waiting for original ProvisioningRequest to apply")
+		waitForPolicies(prBuilder)
+
+		By("waiting for ProvisioningRequest to be fulfilled")
+		prBuilder, err = prBuilder.WaitUntilFulfilled(time.Minute)
+		Expect(err).ToNot(HaveOccurred(), "Failed to wait for ProvisioningRequest to become fulfilled")
+
+		if powerState != "On" {
+			By("waiting for spoke 1 to recover")
```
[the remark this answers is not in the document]
klaskosk: this is for the day2 hardware profile upgrade test, 77391 (the last one). admittedly it's subject to change once we try out the DTIAS

tests/cnf/ran/oran/internal/tsparams/consts.go
```diff
@@ -31,6 +31,10 @@ const (
 	HugePagesSizeKey = "hugepages-size"
 	// OCloudSiteID is the name of the site in the hardware manager to provision in.
 	OCloudSiteID = "rdu3"
+	// PolicySelectorLabel is the ExtraLabel applied to the managed cluster that determines which policies to apply.
+	PolicySelectorLabel = "sno-ran-du-policy"
```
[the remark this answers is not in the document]
klaskosk: this label is part of the ClusterInstance defaults ConfigMap (which is immutable) so we shouldn't have to worry about deleting it accidentally. The name just follows what the devs used when they made the original example git config but ultimately it's arbitrary as long as it matches the ZTP git repo

tests/cnf/ran/oran/tests/oran-post-provision.go
klaskosk: Looking back yeah this is pretty much identical, it's entirely in the git configs really. The template version changes both the policyTemplateParameters schema but also the policyTemplate defaults ConfigMap so we don't update the ProvisioningRequest but still observe the changes

At minimum I'll add some more context in comments so it's clearer what should be happening

tests/cnf/ran/oran/tests/oran-post-provision.go
klaskosk: yes, it's managed entirely by the policies

klaskosk: wait I'm wrong here, it gets deleted in BeforeEach. I'll move that to AfterEach since like with the labels it's really a clean up step

tests/cnf/ran/oran/tests/oran-post-provision.go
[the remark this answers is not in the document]
yliu127: okay

yliu127 (approved):

PR #463, opened 2025-02-22
ci: run eco-goinfra bump for all branches

klaskosk:
This PR updates the existing eco-goinfra-bump to take a branch input and remove the schedule. Instead, a new workflow has been added on a schedule that gets all branches on both eco-gotests and eco-goinfra and runs the bump against branches that exist on both. Since the setup-go action caches automatically, runs against branches with no need to bump will be fast (<30 seconds).

[Example run](https://github.com/klaskosk/eco-gotests/actions/runs/13467685849/job/37636771757), the failure is just because all the reviewers aren't members of my fork

### DIFF
--- a/.github/workflows/bump-all-branches.yml
+++ b/.github/workflows/bump-all-branches.yml
@@ -1,0 +1,51 @@
+name: Bump eco-goinfra on all branches
+on:
+  workflow_dispatch:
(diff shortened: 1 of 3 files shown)


klaskosk:
Closing in favor of #1211

PR #467, opened 2025-02-23
cnf ran: fix small post-provision automation issue

klaskosk:
There are a few lingering automation issues caused by race conditions between the tests and the operator on the cluster. This PR addresses them in two ways:

1. Adds a sleep to 77374. This is a bit of a workaround but for now the simplest way to ensure that we do not end up in the situation where a policy template is updated but its compliance state refers to the old template.
1. Adds an additional step to waitForPolicies which ensures that the policy details in the ProvisioningRequest status match the desired policy version. Previously, we could end up in a situation where all the policies have propagated to the spoke but the ProvisioningRequest shows the old policies. This would then cause a 403 as the ProvisioningRequest turns pending after the policies have propagated and the state had been fulfilled.

A more permanent solution would likely be switching to watching resources instead of polling but the current solution requires the least changes for now.

### DIFF
--- a/tests/cnf/ran/oran/internal/helper/helper.go
+++ b/tests/cnf/ran/oran/internal/helper/helper.go
@@ -227,3 +227,29 @@ func WaitForPolicyVersion(client *clients.Settings, namespace, policyVersion str
 			return true, nil
 		})
 }
+
+// WaitForPRPolicyVersion waits up to timeout until all of the policies on the provided ProvisioningRequest have
+// policyVersion.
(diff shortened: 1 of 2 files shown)


klaskosk:
[Tested](https://jenkins-csb-kniqe-auto.dno.corp.redhat.com/job/Private_Folders/job/klaskosk/job/ocp-far-edge-vran-tests/30)

yliu127 (approved):

PR #486, opened 2025-03-06
Renamed e2e IBU test and removed rollback post-upgrade

josclark42:


tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
```diff
@@ -49,44 +49,6 @@ var _ = Describe(
 			})
 		})
 
-		AfterAll(func() {
```
yliu127, line 52: If we just remove this, we are basically loosing test coverage for rollback. 

I suggest we do this: 
- Use Ordered container (without continue on error)
- TC1: upgrade
- TC2: rollback (use this test id for rollback: 69058) - use different test tag, so we can choose to not run this test for debugging purpose.
- Update polarion test case 68954 to remove the teardown to match the actual implementation.

josclark42: @yliu127  You are right that we would lose coverage for rollback after successful upgrade. We do cover rollback in the [rollback after failed upgrade test case](https://github.com/openshift-kni/eco-gotests/blob/main/tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/rollback-after-failed-upgrade-test.go). 

 The goal of this PR was to have the cluster in an upgraded state after the test suite. If we break this test into two test cases as you suggest we will still wind up with a rolled back cluster after the suite finishes. If you are okay with that, I'll make the changes you recommended.

yliu127: yes. but those are two different test cases  - rollback after successful upgrade vs auto rollback on failed upgrade.

yliu127: With the two test cases separated, we can optionally include/exclude rollback after successful upgrade.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -49,7 +49,31 @@ var _ = Describe(
...
+
+			})
+
+			By("Saving target sno cluster info after upgrade", func() {
+				err := cnfclusterinfo.PostUpgradeClusterInfo.SaveClusterInfo()
+				Expect(err).ToNot(HaveOccurred(), "Failed to collect and save target sno cluster info after upgrade")
+			})
+
+		})
+
```
yliu127, line 75: post upgrade validation should be added in the first test. not the second.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -49,7 +49,31 @@ var _ = Describe(
...
+		It("Rollback successful upgrade", reportxml.ID("69058"), func() {
```
yliu127, line 76: Should skip this test if upgrade test failed based on the previous flag.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -49,7 +49,31 @@ var _ = Describe(
...
+				_, err = newIbguBuilder.WaitUntilComplete(30 * time.Minute)
+				Expect(err).NotTo(HaveOccurred(), "Prep and Upgrade IBGU did not complete in time.")
+
+			})
+
+			By("Saving target sno cluster info after upgrade", func() {
+				err := cnfclusterinfo.PostUpgradeClusterInfo.SaveClusterInfo()
+				Expect(err).ToNot(HaveOccurred(), "Failed to collect and save target sno cluster info after upgrade")
+			})
+
```
yliu127, line 73: Need to add a flag here to indicate upgrade passed, which will then be used in the next test as precondition.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
yliu127: @josclark42 just to confirm, is polarion test case 68954  updated with teardown removed?

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -14,10 +15,11 @@ import (
 	"github.com/openshift-kni/eco-gotests/tests/lca/imagebasedupgrade/cnf/internal/cnfclusterinfo"
 	"github.com/openshift-kni/eco-gotests/tests/lca/imagebasedupgrade/cnf/internal/cnfhelper"
 	"github.com/openshift-kni/eco-gotests/tests/lca/imagebasedupgrade/cnf/internal/cnfinittools"
-	cnfibuvalidations "github.com/openshift-kni/eco-gotests/tests/lca/imagebasedupgrade/cnf/internal/validations"
 	"github.com/openshift-kni/eco-gotests/tests/lca/imagebasedupgrade/cnf/upgrade-talm/internal/tsparams"
 )
 
+var upgradeSuccessful bool = false
```
yliu127, line 21: should be under Describe (i.e., define it under the smallest scope possible)

yliu127: Also pls fix the auto generated github action comment. the import should not be moved.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -102,15 +66,58 @@ var _ = Describe(
 				_, err = newIbguBuilder.WaitUntilComplete(30 * time.Minute)
 				Expect(err).NotTo(HaveOccurred(), "Prep and Upgrade IBGU did not complete in time.")
 
+				upgradeSuccessful = true
```
yliu127, line 69: I think it's best to be plugged after row 75, which really indicated a success where we are able to pull the cluster info. Also we might actually need the extracted cluster info in order to run other tests?

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -102,15 +66,58 @@ var _ = Describe(
...
 				Expect(err).ToNot(HaveOccurred(), "Failed to collect and save target sno cluster info after upgrade")
 			})
 
+			By("Running post-upgrade validations", func() {
+				cnfibuvalidations.PostUpgradeValidations()
+			})
 		})
 
-		cnfibuvalidations.PostUpgradeValidations()
+		if upgradeSuccessful {
```
yliu127, line 82: if condition should be added inside the rollback test function, with proper skip message. So it will not be skipped silently.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -102,15 +66,58 @@ var _ = Describe(
...
+					newIbguBuilder := ibgu.NewIbguBuilder(cnfinittools.TargetHubAPIClient,
+						tsparams.IbguName, tsparams.IbguNamespace).
+						WithClusterLabelSelectors(tsparams.ClusterLabelSelector).
+						WithSeedImageRef(cnfinittools.CNFConfig.IbguSeedImage, cnfinittools.CNFConfig.IbguSeedImageVersion).
+						WithOadpContent(cnfinittools.CNFConfig.IbguOadpCmName, cnfinittools.CNFConfig.IbguOadpCmNamespace).
+						WithPlan([]string{"Prep", "Upgrade"}, 5, 30)
+
+					_, err := newIbguBuilder.DeleteAndWait(1 * time.Minute)
+					Expect(err).ToNot(HaveOccurred(), "Failed to delete prep-upgrade ibgu on target hub cluster")
+
```
yliu127, line 94: Should have separate step logging for creating rollback IBGU?

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -102,15 +66,58 @@ var _ = Describe(
...
+						cnfinittools.CNFConfig.IbguOadpCmNamespace)
+					rollbackIbguBuilder = rollbackIbguBuilder.WithPlan([]string{"Rollback", "FinalizeRollback"}, 5, 30)
+
+					rollbackIbguBuilder, err = rollbackIbguBuilder.Create()
+					Expect(err).ToNot(HaveOccurred(), "Failed to create rollback Ibgu.")
+
+					_, err = rollbackIbguBuilder.WaitUntilComplete(30 * time.Minute)
+					Expect(err).NotTo(HaveOccurred(), "Rollback IBGU did not complete in time.")
+
+					_, err = rollbackIbguBuilder.DeleteAndWait(1 * time.Minute)
```
yliu127, line 111: Just curious - why are we deleting the rollback ibgu at the end? 
Also there should be a step logging?

josclark42: @yliu127 We delete the rollback ibgu at the end of the test so that the target env is in the same state it was in prior to the test. This way, if we need to re-run the test suite we can do so without having to manually remove the rollback IBGU.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
josclark42: Fixed- thank you.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
josclark42: Done and done. thank you.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
josclark42: Added skip test if !upgradeSuccessful

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -49,7 +50,40 @@ var _ = Describe(
...
+			upgradeSuccessful = true
+
+			By("Running post-upgrade validations", func() {
+				cnfibuvalidations.PostUpgradeValidations()
+			})
+		})
+
+		It("Rollback successful upgrade", reportxml.ID("69058"), func() {
+			if !upgradeSuccessful {
+				Skip("Skipping post-upgrade validations due to upgrade failure")
```
yliu127, line 84: Skipping rollback test due to upgrade failure.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
josclark42: @yliu127 I revised the test case by setting it to 'unapproved' first, as you suggested.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -49,7 +50,40 @@ var _ = Describe(
...
+			})
+
+			By("Saving target sno cluster info after upgrade", func() {
+				err := cnfclusterinfo.PostUpgradeClusterInfo.SaveClusterInfo()
+				Expect(err).ToNot(HaveOccurred(), "Failed to collect and save target sno cluster info after upgrade")
+			})
+
+			upgradeSuccessful = true
+
+			By("Running post-upgrade validations", func() {
```
yliu127, line 77: I think this By should be removed. cnfibuvalidations.PostUpgradeValidations() is not a test step, they are test cases being invoked in this test (which is confusing..).

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/e2e-upgrade-test.go
```diff
@@ -25,7 +25,8 @@ var _ = Describe(
 	Label(tsparams.LabelEndToEndUpgrade), func() {
 
 		var (
-			clusterList []*clients.Settings
+			clusterList       []*clients.Settings
+			upgradeSuccessful = false
```
yliu127, line 29: suggest to initiate this as true (and set it to false when we enter the first test case), so that it is possible to run rollback test case on a upgraded cluster as a standalone test - could be helpful in failure analysis/rerun from specific stage.

yliu127 (approved):

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
```diff
@@ -49,7 +50,40 @@ var _ = Describe(
...
+			})
+
+			By("Saving target sno cluster info after upgrade", func() {
+				err := cnfclusterinfo.PostUpgradeClusterInfo.SaveClusterInfo()
+				Expect(err).ToNot(HaveOccurred(), "Failed to collect and save target sno cluster info after upgrade")
+			})
+
+			upgradeSuccessful = true
+
+			cnfibuvalidations.PostUpgradeValidations()
```
klaskosk, line 79: this should be outside of the `It` since it actually contains a `Describe` node. ST tends to write ginkgo tests outside of the main test file and then just use a function to include them

yliu127: Suggest we let this go and refactor later as this is how it was written originally, and the main goal for this PR is to split upgrade and rollback into two different test cases so that we can leave the cluster in upgraded state to investigate post upgrade validation test failure further. 

The complication for taking out the post validation test cases I think is in "Continue On Failure", while we want post validation test cases to continue on failure, we also want the first upgrade test to succeed before running the rest of the test cases, so there are some thoughts we need to put into this and refactor.

yliu127: Ah I see what you mean... @josclark42  ^^ can you pls make the change and test it?

josclark42: Updated- thank you both.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
```diff
@@ -49,7 +50,43 @@ var _ = Describe(
...
+
+			upgradeSuccessful = true
+
+		})
+
+		if upgradeSuccessful {
+			cnfibuvalidations.PostUpgradeValidations()
+		}
+
+		It("Rollback successful upgrade", reportxml.ID("69058"), func() {
```
yliu127, line 85: @klaskosk what would you recommend if we want to disable this test case in CI profile via vran far edge job? Shall we add a label for this test case or we can exclude this by polarion test id?

klaskosk: I'd probably use the ID directly since you get into label sets due to the colon which work differently

yliu127: great!

yliu127: CC @josclark42  ^^

yliu127 (dismissed):

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
```diff
@@ -2,6 +2,7 @@ package upgrade_test
 
 import (
 	"fmt"
+	cnfibuvalidations "github.com/openshift-kni/eco-gotests/tests/lca/imagebasedupgrade/cnf/internal/validations"
```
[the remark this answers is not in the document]
josclark42: done

yliu127 (approved):

klaskosk (approved):
/lgtm

PR #490, opened 2025-03-10
ci: backport eco-goinfra's fix to prevent forks from running nightly crons

cdvultur:
Applied same fix for from: https://github.com/openshift-kni/eco-goinfra/pull/961 

This will prevent running the actions that will surely fail.

### DIFF
--- a/.github/workflows/docker-image-manual.yml
+++ b/.github/workflows/docker-image-manual.yml
@@ -10,6 +10,7 @@ on:
 jobs:
   build:
     name: Build and push docker image
+    if: github.repository_owner == 'openshift-kni'
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/docker-image-ci.yml
+++ b/.github/workflows/docker-image-ci.yml
@@ -12,6 +12,7 @@ on:
 jobs:
   build:
     name: Build and push docker image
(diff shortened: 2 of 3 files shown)


.github/workflows/docker-image-manual.yml
```diff
@@ -10,6 +10,7 @@ on:
 jobs:
   build:
     name: Build and push docker image
+    if: github.repository_owner == 'openshift-kni'
```
klaskosk, line 13: since this has to be triggered manually it probably doesn't make a difference but can't hurt

klaskosk (approved):
/lgtm

.github/workflows/docker-image-manual.yml
cdvultur: Even if ran manually, it will fail because of https://github.com/openshift-kni/eco-gotests/blob/main/.github/workflows/docker-image-manual.yml#L38 

I think that is supposed to be ran by hand in `openshift-kni` ?

klaskosk: true you're right, it won't cause all the automatic email spam but this should prevent it being accidentally ran and failing

sebrandon1 (approved):

PR #500, opened 2025-03-14
yaml: Address mismatch of v2/v3 imports

sebrandon1:
Switch totally over to using `yaml.v3`.


klaskosk (dismissed):
a little nervous about the major version change but looking at the differences I don't think anyone was intentionally using v2 for anything

PR #502, opened 2025-03-14
metallb: use service.GetGVR()

sebrandon1:
Follow up to #209 

#209 was created way before these tests referenced the old deprecated func.


klaskosk (approved):
/lgtm
cc: @gkopels @evgenLevin

PR #505, opened 2025-03-20
System Test: RDS Core - pod-level bond test suite

elenagerman:
pod-level bond test suite


elenagerman:
@yprokule PTAL, it's ready for review

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+	scheduleOnHost,
+	sriovNet1Name,
+	sriovNet2Name,
+	bondNetName,
+	bondInfIPv4,
+	bondInfIPv6,
+	bondInfSubMaskIPv4,
+	bondInfSubMaskIPv6,
+	bondInfMacAddr string) error {
+	glog.V(100).Infof("Creating the pod-level bonded deployment test source deployment")
```
yprokule, line 55: The next line seems to cleanup the deployment:
```
err := cleanUpPodLevelBondDeployment(apiClient, deploymentName, nsName, podLabel)
```

so message is a bit misleading
```suggestion
	glog.V(100).Infof("Ensuring deployment %q doesn't exist in %q namespace", deploymentName, nsName)
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+			deploymentName, nsName, err)
+
+		return fmt.Errorf("failed to cleanup deployment %s from the namespace %s: %w",
+			deploymentName, nsName, err)
+	}
+
+	glog.V(100).Infof("Removing ServiceAccount")
+	deleteServiceAccount(podLevelBondDeploymentSAName, nsName)
+
+	glog.V(100).Infof("Creating ServiceAccount")
```
yprokule, line 69: ditto
```suggestion
	glog.V(100).Infof("Creating ServiceAccount %q", podLevelBondDeploymentSAName)
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+	createServiceAccount(podLevelBondDeploymentSAName, nsName)
+
+	glog.V(100).Infof("Removing Cluster RBAC")
+	deleteClusterRBAC(podLevelBondDeploymentRBACName)
+
+	glog.V(100).Infof("Creating Cluster RBAC")
```
yprokule, line 75: ditto

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+
+	glog.V(100).Infof("Defining container configuration")
+
+	deploymentContainer := definePodLevelBondDeploymentContainer()
+
+	glog.V(100).Infof("Obtaining container definition")
+
+	deployContainerCfg, err := deploymentContainer.GetContainerCfg()
+	if err != nil {
+		return fmt.Errorf("failed to obtain container definition: %w", err)
```
yprokule, line 90: please log there was an error retrieving container's config

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+	}
+
+	glog.V(100).Infof("Defining deployment configuration")
```
yprokule, line 93: ditto, please log deployment's name

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+		},
+	}
+
+	podDeployment := deployment.NewBuilder(apiClient, deploymentName, nsName, deployLabels, *containerConfig)
+
+	glog.V(100).Infof("Assigning ServiceAccount %q to the deployment", podLevelBondDeploymentSAName)
+
+	podDeployment = podDeployment.WithServiceAccountName(podLevelBondDeploymentSAName)
+
+	podDeployment = podDeployment.WithNodeSelector(nodeSelector)
```
yprokule, line 254: similar to logging assignment of SA to the deployment I'd log setting _nodeSelector_
```suggestion
        glog.V(100).Infof("Assigning NodeSelector %q to the deployment", nodeSelector)
	podDeployment = podDeployment.WithNodeSelector(nodeSelector)
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+		},
+		{
+			Name:      sriovNet2Name,
+			Namespace: nsName,
+		},
+		{
+			Name:      bondNetName,
+			Namespace: nsName,
+			IPRequest: []string{fmt.Sprintf("%s/%s", bondInfIPv4, bondInfSubMaskIPv4),
+				fmt.Sprintf("%s/%s", bondInfIPv6, bondInfSubMaskIPv6)},
```
yprokule, line 243: It worth validating input, for e.g. checking that both provided params ( IP address and subnet) ain't empty.
Also what about a case when only IPv4 or IPv6 pair is passed?

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+	if err != nil {
+		glog.V(100).Infof("Failed to wait for pod %q in namespace %q to become Ready: %v",
+			clientPod.Definition.Name, clientPod.Definition.Namespace, err)
+
+		return "", fmt.Errorf("failed to wait for pod %q in namespace %q to become Ready: %w",
+			clientPod.Definition.Name, clientPod.Definition.Namespace, err)
+	}
+
+	cmdToRun := []string{"bash", "-c",
+		fmt.Sprintf("testcmd -protocol tcp -port %s -interface net3 -server %s",
```
yprokule, line 278: do U think it worth parameterizing protocol ?

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+				clientPod.Object.Name, clientPod.Definition.Namespace)
+
+			output = result.String()
+			glog.V(100).Infof("Command's output:\n\t%v", output)
+
+			return true, nil
+		})
+
+	if err != nil {
+		return "", fmt.Errorf("failed to run command from within pod %s: %w", clientPod.Object.Name, err)
```
yprokule, line 310: please log there was an error while running the command

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+			serverPort, packetsNumber, timeout, serverIPAddr)}
+
+	glog.V(100).Infof("Execute command: %q", cmdToRun)
+
+	var output string
+
+	err = wait.PollUntilContextTimeout(
+		context.TODO(),
+		time.Second*3,
+		time.Minute,
```
yprokule, line 348: how does this work with the `-packages` and `-timeoutTCP` params from the _cmdToRun_ ?

For e.g. if one specifies `-timeoutTCP 10min`, will this _wait.PollUntilContextTimeout_ work?

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+		})
+
+	if err != nil {
+		return "", fmt.Errorf("failed to run command from within pod %s: %w", clientPod.Object.Name, err)
+	}
+
+	return output, nil
+}
+
+func generateLongTCPTraffic(
```
yprokule, line 316: The only difference from `generateTCPTraffic` I spotted are params of the _cmdToRun_ variable.
Does it make sense to add extra params to the aforementioned function. to avoid code duplication?

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+func findInCmdExecOutput(cmdExecOutput, stringToFind string) (bool, error) {
+	var err error
+
+	matchesFound := 0
+
+	if len(cmdExecOutput) != 0 {
+		buf := new(bytes.Buffer)
+		_, err = buf.WriteString(cmdExecOutput)
+
+		stringToFind, _ := regexp.Compile(stringToFind)
```
yprokule, line 385:
```suggestion
		stringToFind, err := regexp.Compile(stringToFind)
```

yprokule: if there's a check for writing to buffer(based on the following messages) then this `regexp.Compile` call should be done after error's evaluation

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+				matchesFound++
+			}
+		}
+
+		if matchesFound < 1 {
+			glog.V(100).Infof("Expected string %q not found in the output: %s", stringToFind, cmdExecOutput)
+
+			return false, fmt.Errorf("expected string %q not found in the output : %s", stringToFind, cmdExecOutput)
+		}
+	}
```
yprokule, line 407: use case `len(cmdExecOutput) == 0` not covered

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+
+func getBondActiveInterface(clientPod *pod.Builder) (string, error) {
+	glog.V(90).Infof("Getting bond active VF interface for the pod %s in namespace %s",
+		clientPod.Definition.Name, clientPod.Definition.Namespace)
+
+	var output bytes.Buffer
+
+	var result string
+
+	var err error
```
yprokule, line 440: nit, this could be grouped like:
```suggestion
	var (
	    output bytes.Buffer
            result   string
            err        error
         )
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+
+	if toDisable {
+		expectedInfState = "down"
+	} else {
+		expectedInfState = "up"
+	}
+
+	glog.V(100).Infof("Disable pod-level bond interface %s for the pod %s in namespace %s",
```
yprokule, line 549: depending on the `toDisable` value it could be either _Enable_ or _DIsable_

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+				ipv4Addr, podObj.Definition.Name, podObj.Definition.Namespace, output)
+
+			return false, fmt.Errorf("ipv4 address %s not found configured for the bond interface "+
+				"of the pod %s in namespace %s: %s",
+				ipv4Addr, podObj.Definition.Name, podObj.Definition.Namespace, output)
+		}
+	}
+
+	if ipv6Addr != "" {
+		glog.V(100).Infof("Ensure ipv6 %s address defined as expected", ipv6Addr)
```
yprokule, line 653:
```suggestion
		glog.V(100).Infof("Ensure IPv6 %s address defined as expected", ipv6Addr)
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+	Expect(err).ToNot(HaveOccurred(),
+		fmt.Sprintf("Failed to retrieve client pod-level bond %s object from namespace %s: %v",
+			clientDeploymentName, clientDeploymentNamespace, err))
+
+	serverPodObj, err := getPodObjectByNamePattern(APIClient, serverDeploymentName, serverDeploymentNamespace)
+	Expect(err).ToNot(HaveOccurred(),
+		fmt.Sprintf("Failed to retrieve server pod-level bond %s object from namespace %s: %v",
+			serverDeploymentName, serverDeploymentNamespace, err))
+
+	By("Inspect pod-level bonded interface for the client pod")
```
yprokule, line 741:
```suggestion
	By("Inspecting bonded interface within the client pod")
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+
+	isFound, err := inspectPodLevelBondedInterfaceConfig(clientPodObj, clientIPv4, clientIPv6)
+	Expect(err).ToNot(HaveOccurred(),
+		fmt.Sprintf("Failed to verify client pod-level bond pod %s from namespace %s config: %v",
+			clientPodObj.Definition.Name, clientPodObj.Definition.Namespace, err))
+	Expect(isFound).To(Equal(true),
+		fmt.Sprintf("The pod-level bonded interface for the pod %s in namespace %s not as expected",
+			clientPodObj.Definition.Name, clientPodObj.Definition.Namespace))
+
+	By("Inspect pod-level bonded interface for the server pod")
```
yprokule, line 751:
```suggestion
	By("Inspecting bonded interface within the server pod")
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+		schedulerOnHost = RDSCoreConfig.PodLevelBondPodOneScheduleOnHost
+	}
+
+	netOne := RDSCoreConfig.PodLevelBondSRIOVNetOne
+	netTwo := RDSCoreConfig.PodLevelBondSRIOVNetTwo
+
+	if !samePF {
+		netOne = RDSCoreConfig.PodLevelBondSRIOVNetTwo
+		netTwo = RDSCoreConfig.PodLevelBondSRIOVNetOne
+	}
```
yprokule, line 845: could U please explain this part? if _samePF_ is _false_ then _netOne_ shouldn't use same PF as _netTwo_. But here it swaps values from _netTwo_ to _netOne_

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
elenagerman: I don't think so, for what? this is internal method and we have to use reliable protocol. for the testcmd tool we have only 4 options: icmp, udp, tcp or sctp. icmp&udp not reliable, sctp not required in our case

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
elenagerman: -timeoutTCP is not a 10 min, but 10 sec.
tcp timeout meaning is a **connection** TCP timeout (how long to retransmit and leave TCP session open when no response from the other side), recommendations are different, the testcmd app default is 2 sec.

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1114 @@
...
+
+	isFound, err = inspectPodLevelBondedInterfaceConfig(serverPodObj, serverIPv4, serverIPv6)
+	Expect(err).ToNot(HaveOccurred(),
+		fmt.Sprintf("Failed to verify server pod-level bond pod %s from namespace %s config: %v",
+			serverPodObj.Definition.Name, serverPodObj.Definition.Namespace, err))
+	Expect(isFound).To(Equal(true),
+		fmt.Sprintf("The pod-level bonded interface for the pod %s in namespace %s not as expected",
+			serverPodObj.Definition.Name, serverPodObj.Definition.Namespace))
+
+	if serverIPv4 != "" {
```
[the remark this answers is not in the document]
elenagerman: it is specifically defined this way. in case there is a need to skip one of the verifications. therefore, the method failure on an empty ip is not provided

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
elenagerman: I'm sorry, I'm not so clearly understand your question, but I'll try to explain.
1. regarding code
this part is a default
`
        netOne := RDSCoreConfig.PodLevelBondSRIOVNetOne
	netTwo := RDSCoreConfig.PodLevelBondSRIOVNetTwo
`
which will be changed in case same PF is not required by
`
      if !samePF {
		netOne = RDSCoreConfig.PodLevelBondSRIOVNetTwo
		netTwo = RDSCoreConfig.PodLevelBondSRIOVNetOne
	}
`
2. regarding general config:

- we have two secondary interfaces; just as an example I will take ens3f0 and ens3f1 names
- we have two snnps: left -> ens3f0 & right -> ens3f1
- we have two sriovNetworks:  
                    - bondNetOne -> snnp left -> ens3f0
                    - bondNetTwo -> snnp right -> ens3f1

- we build a bond interfaces based on this networks
          - for the same PF: net1==bondNetOne(snnp left -> ens3f0), net2==bondNetTwo(snnp right -> ens3f1)
          - for the different PFs: net1==bondNetTwo(snnp right -> ens3f1), net2==bondNetOne(snnp left -> ens3f0)
For the bonded interface net1 is always a default after creation. We can change it only if we'll take net1 down (like we do for the failover test)

- there are two way to config different PF for the pod:
                             1. like I did, just to switch networks during bond config and run connectivity check
                             2. create second bond just the with the same config we did for the first bond and take net1 down to switch to the second network (from left to right) and run connectivity check
   Since the second scenario is covered for the different node by the fail-over test, I'm using the first option which is not requires any additional steps will be done after config

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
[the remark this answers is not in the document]
yprokule: what I'm suggesting is to fail the test if both of them are empty

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
yprokule: Could You please explain what does _same PF_ mean in your case?

When I see _same PF_ I expect VFs to be allocated from _same PF_, e.g. `ens3f0` and not one VF from `ens3f0` and another from `ens3f1`

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1197 @@
...
+				glog.V(100).Infof("Error running command from within a pod %q: %v",
+					podObj.Object.Name, err)
+
+				return false, nil
+			}
+
+			glog.V(100).Infof("Successfully executed command from within a pod %q in namespace %q",
+				podObj.Object.Name, podObj.Definition.Namespace)
+
+			output = result.String()
```
yprokule, line 667: while running it in my setup hit an issue that output was empty(more likely some timing issue) cause later IP as there, therefore wdyt about retrying the command if the output it empty?

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1197 @@
...
+
+	serverPodObj, err := getPodObjectByNamePattern(
+		APIClient,
+		RDSCoreConfig.PodLevelBondDeploymentTwoName,
+		RDSCoreConfig.PodLevelBondNamespace)
+	Expect(err).ToNot(HaveOccurred(),
+		fmt.Sprintf("Failed to retrieve server pod-level bond %s object from namespace %s: %v",
+			RDSCoreConfig.PodLevelBondDeploymentOneName, RDSCoreConfig.PodLevelBondNamespace, err))
+
+	activeInf, err := getBondActiveInterface(serverPodObj)
```
yprokule, line 994: I'd add logging like:

```suggestion
        By(fmt.Sprintf("Getting bond's active interface for pod %q", serverPodObj.Definition.name))
	activeInf, err := getBondActiveInterface(serverPodObj)
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1197 @@
...
+	Expect(err).ToNot(HaveOccurred(),
+		fmt.Sprintf("Failed to retrieve bond active interface for the pod deployment %s in namespace %s: %v",
+			serverPodObj.Definition.Name, serverPodObj.Definition.Namespace, err))
+
+	glog.V(100).Infof("DEBUG LOG: active interface found - %s", activeInf)
```
yprokule, line 999:
```suggestion
	glog.V(100).Infof("Active interface found - %s", activeInf)
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1197 @@
...
+
+		testPassed, err := scanClientPodTrafficOutput(output)
+		Expect(err).ToNot(HaveOccurred(),
+			fmt.Sprintf("Failed to parse client pod %s from namespace %s output: %v",
+				clientPodObj.Definition.Name, clientPodObj.Definition.Namespace, err))
+		Expect(testPassed).To(Equal(true),
+			fmt.Sprintf("TCP traffic test verification failed for the pod %s in namespace %s; output %s",
+				clientPodObj.Definition.Name, clientPodObj.Definition.Namespace, output))
+
+		glog.V(100).Infof("DEBUG LOG 222")
```
yprokule, line 1023: please change to smith more meaningful :)

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1197 @@
...
+	Expect(err).ToNot(HaveOccurred(),
+		fmt.Sprintf("Failed to retrieve bond active interface for the pod deployment %s in namespace %s: %v",
+			serverPodObj.Definition.Name, serverPodObj.Definition.Namespace, err))
+
+	glog.V(100).Infof("DEBUG LOG: active interface found - %s", activeInf)
+
+	By("Send data from the client container to the IPv4 address used by the server container")
+
+	go func() {
+		output, err := generateTCPTraffic(
```
yprokule, line 1004:
```suggestion


	go func() {
		By("Send data from the client container to the IPv4 address used by the server container")
		output, err := generateTCPTraffic(
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1197 @@
...
+				serverPodObj.Definition.Name, serverPodObj.Definition.Namespace, err))
+	}()
+
+	var ctx SpecContext
+
+	Eventually(func() bool {
+		newActiveInf, err := getBondActiveInterface(serverPodObj)
+		if err != nil {
+			glog.V(rdscoreparams.RDSCoreLogLevel).Infof(
+				"Failed to retrieve new bond active interface for the pod deployment %s in namespace %s: %v",
```
yprokule, line 1041:
```suggestion
				"Failed to retrieve new bond active interface for the pod %s in namespace %s: %v",
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
```diff
@@ -0,0 +1,1197 @@
...
+				"The bond active interface did not changed yet %q", newActiveInf)
+
+			return false
+		}
+
+		return true
+	}).WithContext(ctx).WithPolling(time.Second).WithTimeout(30*time.Second).Should(BeTrue(),
+		"Fail-Over procedure failure; failed to switch to the new bond active interface")
+
+	glog.V(100).Infof("DEBUG LOG 111")
```
yprokule, line 1060: please change to smth meaningful

tests/system-tests/rdscore/tests/00_validate_top_level.go
```diff
@@ -177,6 +177,38 @@ var _ = Describe(
...
+
+			It("Verifies pod-level bonded workloads after pod bonded interface recovering after failure",
+				Label("pod-level-bond", "pod-level-bond-failure"), reportxml.ID("80489"),
+				rdscorecommon.VerifyPodLevelBondWorkloadsAfterBondInterfaceFailure)
+
+			It("Verifies pod-level bonded workloads after bond interface recovering after both VFs failure",
+				Label("pod-level-bond", "pod-level-bond-failure"), reportxml.ID("80696"),
+				rdscorecommon.VerifyPodLevelBondWorkloadsAfterBothVFsFailure)
+
+			It("Verifies pod-level bonded workloads after pod crashing",
```
yprokule, line 208: I suggest adding `MustPassRepeatedly(3)` to stress SR-IOV operator a bit
```suggestion
			It("Verifies pod-level bonded workloads after pod crashing",
			    MustPassRepeatedly(3),
```

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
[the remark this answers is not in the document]
elenagerman: good idea! give me a few mins to modify it

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
elenagerman: this is exactly the meaning :) same PF == both VFs are on the same PF=ens3f0(for example)

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
elenagerman: let me check; if you ran it on jenkins, please share this failure to see exact the scenario

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
elenagerman: sure

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
elenagerman: ups, sorry, it was debug stage left-overs. I remove this like we have this info printed-out inside  getBondActiveInterface method

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
elenagerman: sorry, leftovers, removing :)

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
elenagerman: I did not succeed to reproduce such a failure, but I added a verification for the empty output, I believe it should solve the problem. let me know if you see it does not, thank you!

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
yprokule: Both of them are passed to the _createPrivilegedPodLevelBondDeployment_ method:
```
	err := createPrivilegedPodLevelBondDeployment(
 		APIClient,
 		RDSCoreConfig.PodLevelBondDeploymentTwoName,
 		RDSCoreConfig.PodLevelBondNamespace,
 		podLevelBondPodLabel,
 		schedulerOnHost,
 		netOne,       <<< HERE
 		netTwo,       <<< HERE
 		podLevelBondNetName,
 		RDSCoreConfig.PodLevelBondDeploymentTwoIPv4,
 		RDSCoreConfig.PodLevelBondDeploymentTwoIPv6,
 		RDSCoreConfig.PodLevelBondPodSubnetMaskIPv4,
 		RDSCoreConfig.PodLevelBondPodSubnetMaskIPv6,
 		RDSCoreConfig.PodLevelBondPodMacAddr)
```
and _netOne_ and _netTwo_ still point to 2 different PFs ( _RDSCoreConfig.PodLevelBondSRIOVNetTwo_ and _RDSCoreConfig.PodLevelBondSRIOVNetOne_ ) so how are they from the same PF (unless _RDSCoreConfig.PodLevelBondSRIOVNetTwo == RDSCoreConfig.PodLevelBondSRIOVNetOne_)?

cgoncalves:
Shouldn't the `miimon` parameter in the bond NAD also be set?
I remember it not being set in the bond unless specified by the user, which caused the bond not to detect when the active slave goes down and fail over to the standby slave.

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
elenagerman: Pod level bonding feature == bonding two interfaces from the different vfs/pfs.
So it is ok that sriov bonding pod config includes both (privileged or not privileged).
To create a pod-level bonding pod, we have to provide information about both networks to config interfaces, and one of them will be used as an 'active' for the bond.
This pod will have:
inf net1 = sriovNetOne
inf net2 = sriovNetTwo
inf net3 == bond, with will use only one of the net1&net2 infs as an active
during bond inf creation. It caches the first one (net1) by default. The second one is not in use until something happens with the first one.

We create one such a pod (not privileged) during our day2 config (ocp-edge, metallb app stage) -> we have a pod with the bonded interface net3 when the active interface is net1 == sriovNetOne == VF on PF one (ens3f0 in my case)
We create another one (privileged) pod during the test execution itself -> we have pod with the bonded interface net3 when the active interface is net1 == sriovNetOne == VF on PF one (ens3f0 in my case)
In case on one of the pods (we do it on the second one) we simulate net1 interface failure (take it down, failover scenario), so we switch to the the net2 interface to be active: net2 == sriovNetTwo == VF on PF two (ens3f1 in my case) == different VFs&PFs scenario
Note: We have to have one of the pods privileged to be possible to manipulate with the net1, net2, and net3 pod interfaces states (down/up)

elenagerman:
> Shouldn't the `miimon` parameter in the bond NAD also be set? I remember it not being set in the bond unless specified by the user, which caused the bond not to detect when the active slave goes down and fail over to the standby slave.

It is configured, it is a part of the NAD configuration:
`
      - path: source-crs/optional/networking/networkAttachmentDefinition.yaml
        patches:
          - metadata:
              name: bond-net
              namespace: rds-bonded-sriov-wlkd
            spec:
              config: '{ "type": "bond", 
                "cniVersion": "0.3.1", 
                "name": "bond-net", 
                "mode": "active-backup", 
                "failOverMac": 1, 
                "linksInContainer": true, 
                "miimon": "100", 
                "mtu": 9000, 
                "links": [ {"name": "net1"}, {"name": "net2"} ], 
                "ipam": { "type": "static",
                  "routes": [{"dst": "10.18.93.64/26", "dst": "2620:52:0:15d::40/122"}] } }'

`

elenagerman:
@yprokule can we finalize our review? do you see any other code issue need to be fixed?

tests/system-tests/rdscore/internal/rdscorecommon/sriov-pod-level-bond.go
yprokule: If I understand it right- U want to deploy a 2nd pod that will have active link on the same(or different) PF as the 1st one, that's correct?
If that's correct:
* U need to check what link is active on the 1st pod, and not assume it's always the one referenced by the _net1_, and based on it pickup same/different PF for 2nd pod
* please add a doc string that explains that _samePF_ is relative to the active link of the pod X

yprokule (approved):
/lgtm, thanks

PR #508, opened 2025-03-20
system-test: dedicated parameter for FRR namespace and node filtering

yprokule:


elenagerman:
@yprokule don't miss this one as well: https://github.com/openshift-kni/eco-gotests/blob/main/tests/system-tests/rdscore/internal/rdscorecommon/metallb-graceful-restart.go#L548

PR #510, opened 2025-03-24
system-test: Filter FRR nodes per node

yprokule:
FRR pods now run on all the nodes so we need an option to filter them by node


elenagerman:
@yprokule can we merge it?

PR #540, opened 2025-04-09
system-tests: retry logs retrieval until match found

yprokule:
Fix inaccuracy in the _ScanTCPDumpPodLogs_
Current flow is:
- retrieve log
- if it's empty retry
- if it's not empty - search for the matching string

Proposed change is:
- retrieve log
- if it's empty - retry
- if it's not empty - search for the matching string
- if match found - success
- if no match found - retrieve newer log and search in the newer content


elenagerman:
Enhancements are always good!! 

/lgtm

PR #565, opened 2025-04-18
cnf ran: skip powersave tests on arm

klaskosk:
Depends-on: openshift-kni/eco-goinfra#1004

Changing power states is not currently supported on ARM and can be covered by the compute team, so this PR updates the tests to check for the node architecture and skip if it is not amd64. Additionally, it changes the failure on non-SNO clusters to be a skip.

Additionally, the CPU frequency tests have been modified slightly to avoid deleting and recreating the performance profile. Some other minor changes were made to avoid hitting the funlen lint.

Finally, the version of the cluster-node-tuning operator dependency has been increased to avoid the issue with hugepage sizes on aarch64.


klaskosk:
Tested: https://jenkins-csb-kniqe-auto.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/404

PR #576, opened 2025-04-25
cnf ran: fix cleanup issue in talm precache

klaskosk:
Test case `48902 Tests image precaching - operators` fails to cleanup the CGU it creates, causing issues with other test cases. This PR addresses it by deleting the CGU directly. The rest of the resources will be suffixed and should already be deleted.


klaskosk:
Tested: https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/3021
The failure is a product issue but without this PR it will cause a second test to fail too

yliu127 (approved):

PR #582, opened 2025-04-30
cnf network: fix/improve metallb tests

ajaggapa:
- frrk8s webhook server starts after deploying metallb instance for the first on the cluster. 1st test will fail if eco-gotests is ran right after cluster deployment. Hence, adding wait time.
- ReportIfFailed does not have correct test namespace and relevant CR/Objects to be fetched
- frrk8s test creates metallb instance twice
- use workerNodeList var instead cnfWorkerNodeList while fetching frrk8s pods. Tests fail on 3 worker node setup with this.
- remove using  cnfWorkerNodeList var everywhere else in metallb tests and use workerNodeList var only
- addlabel to the node will attempt to remove it first. This is to avoid test failures when we are trying to add a label to the node which already has the same label.

### DIFF
--- a/tests/cnf/core/network/metallb/internal/metallbenv/metallbenv.go
+++ b/tests/cnf/core/network/metallb/internal/metallbenv/metallbenv.go
@@ -118,7 +118,20 @@ func CreateNewMetalLbDaemonSetAndWaitUntilItsRunning(timeout time.Duration, node
 	glog.V(90).Infof("Verifying if the FRR webhook server is deployed and ready")
 
 	// Check FRR Webhook Server Readiness **(LAST STEP)**
-	frrk8sWebhookDeployment, err := deployment.Pull(APIClient, tsparams.FrrK8WebHookServer, NetConfig.Frrk8sNamespace)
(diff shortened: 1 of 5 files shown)


tests/cnf/core/network/metallb/internal/metallbenv/metallbenv.go
```diff
@@ -118,7 +118,20 @@ func CreateNewMetalLbDaemonSetAndWaitUntilItsRunning(timeout time.Duration, node
...
+					tsparams.FrrK8WebHookServer, NetConfig.Frrk8sNamespace)
+
+				return false, nil
+			}
+
+			return true, nil
+		})
 	if err != nil {
 		return fmt.Errorf("failed to pull the frrk8s webhook server: %w", err)
 	}
```
evgenLevin, line 137: please remove

evgenLevin (dismissed):
/lgtm

evgenLevin (approved):

kononovn (approved):
/lgtm

PR #585, opened 2025-05-01
IBU: Refactored ValidateNoImagesPulled test

josclark42:
This change prevents the test from failing when  $excludeImages var is empty. 

Before this change, the test will fail with the error:

```
 [FAILED] Images are being pulled after upgrade: Usage: grep [OPTION]... PATTERNS [FILE]...
  Try 'grep --help' for more information.

  Expected
      <int>: 83
  not to be >
      <int>: 1
  In [It] at: /home/testuser/tests/lca/imagebasedupgrade/cnf/internal/validations/post_ibu_validations.go:352
```

Because when  $excludeImages var is empty, the grep command fails and isses an an error with character count=83

### DIFF
--- a/tests/lca/imagebasedupgrade/cnf/internal/validations/post_ibu_validations.go
+++ b/tests/lca/imagebasedupgrade/cnf/internal/validations/post_ibu_validations.go
@@ -336,12 +336,9 @@ func ValidateNoImagesPulled() {
 			catSrc, err := olm.ListCatalogSources(TargetSNOAPIClient, "openshift-marketplace")
 			Expect(err).ToNot(HaveOccurred(), "Failed to list catalog sources")
 
-			excludeImages := "| grep -v "
-			for i, catalogSource := range catSrc {
-				excludeImages += catalogSource.Definition.Spec.Image
-				if i < len(catSrc)-1 {
-					excludeImages += " | grep -v "
-				}
+			excludeImages := ""
+			for _, catalogSource := range catSrc {
+				excludeImages += fmt.Sprintf("|grep -v %s", catalogSource.Definition.Spec.Image)
 			}
 
 			logCmd := "journalctl -l --system | grep Pulling | grep image:"


klaskosk (approved):

tests/lca/imagebasedupgrade/cnf/internal/validations/post_ibu_validations.go
```diff
@@ -336,12 +336,9 @@ func ValidateNoImagesPulled() {
...
 
-			excludeImages := "| grep -v "
-			for i, catalogSource := range catSrc {
-				excludeImages += catalogSource.Definition.Spec.Image
-				if i < len(catSrc)-1 {
-					excludeImages += " | grep -v "
-				}
+			excludeImages := ""
+			for _, catalogSource := range catSrc {
+				excludeImages += fmt.Sprintf("|grep -v %s", catalogSource.Definition.Spec.Image)
```
yliu127, line 341: we changed the logic of original PR. Pls add the if condition back as the intent of original PR seems to ignore the last catsrc for whatever reason.

klaskosk: @yliu127 the original logic didn't ignore the last catalogsource, it was just written oddly
```go
// let len(catSrc) == 2
excludedImages := "| grep -v"
// at i = 0
excludedImages ==  "| grep -v zerothImage"
// inside if condition since 0 < len(catSrc)-1
excludedImages == "| grep -v zerothImage | grep -v"
// at i = 1 = len(catSrc) - 1
excludedImages == "| grep -v zerothImage | grep -v lastImage"
// if statement does not execute since i == len(catSrc) - 1
```

The logic does change for `len(catSrc) == 0` since `excludedImages == "| grep -v"` whereas this PR changes it to `excludedImages == ""`. They behave identically for `len(catSrc) >= 1`

josclark42: I don't think the original version ignores the last catsrc. It's specifying `if i < len(catSrc)-1` because i starts at 0, and len() starts at 1. for example, with 3 catalog sources:
```
Iteration 1 (i=0): Adds first image + " | grep -v " (since 0 < 2)
Iteration 2 (i=1): Adds second image + " | grep -v " (since 1 < 2)
Iteration 3 (i=2): Adds third image only (since 2 is not < 2)
```

klaskosk (approved):

yliu127 (approved):

PR #586, opened 2025-05-02
cnf ran: update oran automation to use o2ims api

klaskosk:
This switches the cnf/ran/oran tests to use the O2IMS API rather than the Kubernetes API when interacting with the ProvisioningRequests.

There are a few changes necessary to support this even if the ProvisioningRequestBuilder is still used. Notably, since the only form of status provided is in the ProvisioningPhase, the tests must be changed to not rely on the conditions and policy data. This is achieved through switching to the WaitForPhaseAfter method which allows us to ensure that the changes to the phase occur after changes are made to the ProvisioningRequest.

This also updates the eco-goinfra version to include the O2IMS changes.


klaskosk:
Tested: https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/3036/
This was run a while ago but all that has changed is rebasing, not the code itself

yliu127 (approved):
/lgtm

PR #597, opened 2025-05-08
system-tests: day2 configuration using skopeo

Demostenes777:
This PR changes the way to upgrade/downgrade operators from podman to skopeo and changes path to files to facilitate the mount of volumes when running the tests from the container.


mcornea:
lgtm

PR #605, opened 2025-05-14
System Test: clean-up for unhealthy dpdk and pod-level bond pods

elenagerman:
clean-up for unhealthy dpdk and pod-level bond pods


elenagerman:
@yprokule PTAL

PR #610, opened 2025-05-16
deps: change mergo replace to vanity url

klaskosk:
mergo v1 switched from the github URL to the vanity URL of dario.cat/mergo. This PR changes the replace directive to use the new vanity URL for mergo, eliminating any fragmentation that could result from other dependencies requiring mergo by its vanity URL.


klaskosk:
Now that we have an indirect dependency on `dario.cat/mergo@v1.0.1` this PR won't work. Probably for the better if it was going to end up being an issue

PR #615, opened 2025-05-21
cnf network: Remove unnecessary GW mode reversion calls

evgenLevin:
Removed redundant `setLocalGWMode(false)` calls from tests as they are no longer required. This simplifies the teardown process and ensures cleaner test logic.


ajaggapa (approved):
lgtm

gkopels (approved):

PR #616, opened 2025-05-22
cnf ran: fix automation issue due to time ambiguity

klaskosk:
O-RAN post-provision test case 77375 regularly fails since it can complete within a single second and WaitForPhaseAfter produces false negatives in this situation. Since the precision of times on the cluster cannot be changed, this PR solves the issue by waiting to ensure the ProvisioningRequest will not be updated in the same second as the start time.

[Tested successfully](https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/3232)

### DIFF
--- a/tests/cnf/ran/oran/tests/oran-post-provision.go
+++ b/tests/cnf/ran/oran/tests/oran-post-provision.go
@@ -58,7 +58,7 @@ var _ = Describe("ORAN Post-provision Tests", Label(tsparams.LabelPostProvision)
 		prBuilder, err := oran.PullPR(o2imsAPIClient, tsparams.TestPRName)
 		Expect(err).ToNot(HaveOccurred(), "Failed to pull the ProvisioningRequest again")
 
-		restoreTime := time.Now()
+		restoreTime := getStartTime()
 
 		By("restoring the original ProvisioningRequest spec")
 		prBuilder.Definition.Spec = *originalPRSpec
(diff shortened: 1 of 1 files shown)


yliu127 (approved):

tests/cnf/ran/oran/tests/oran-post-provision.go
```diff
@@ -287,3 +287,16 @@ func waitForLabels() {
 	_, err = mcl.WaitForLabel(tsparams.TestName, time.Minute)
 	Expect(err).ToNot(HaveOccurred(), "Failed to wait for spoke 1 ManagedCluster to have the label")
 }
+
+// getStartTime saves the current time, waits until the next second, then returns the saved time. Since Kubernetes only
+// serializes times with second precision, it is impossible to order times within a second. Adding a delay is necessary
+// to ensure WaitForPhaseAfter does not produce false negatives when the ProvisioningRequest transitions from Fulfilled
+// to Pending to Fulfilled within the same second as the call to time.Now().
+func getStartTime() time.Time {
```
dgonyier, line 295: There is a chance that the clusters idea of a second is slightly different from the prov host, so there is a possibility that prov host seconds where the test run are shorter and depending one when the seconds "align", it is possible that this function returns the same second. The fix for this would be adding a small delay to the wait, like 5ms or something to avoid this. I am not sure if `time.Sleep()` is that precise though.

No code changes required. Just something to keep in mind if there are flakes where the same second is returned. In practice, I think this would be an extremely rare event, if it happens at all.

"computer time" is weird... I've had to deal with this kind of thing before.

klaskosk: @dgonyier this is a really good point, I'll keep an eye on it and if it's an issue we can probably just increase the delay to 2 seconds which should give us a fair bit of leeway for the prov host and cluster to have different times

PR #623, opened 2025-05-28
IBU e2e test: moved IBGU cleanup procedure

josclark42:
Moved the cleanup process for the prep->upgrade and rollback IBGUs to an AfterEach() so that they are always cleaned up, even in the event that the upgrade test fails. Previously, when the upgrade failed the IBGU would remain on the hub cluster, causing subsequent tests to fail.

### DIFF
--- a/tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
+++ b/tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
@@ -50,6 +50,21 @@ var _ = Describe(
 			})
 		})
 
+		AfterEach(func() {
+			By("Deleting IBGUs on target hub cluster", func() {
+				_, err := ibgu.NewIbguBuilder(cnfinittools.TargetHubAPIClient,
+					tsparams.IbguName, tsparams.IbguNamespace).DeleteAndWait(1 * time.Minute)
+
+				Expect(err).ToNot(HaveOccurred(), "Failed to delete ibgu on target hub cluster")
+
+			})
+
+			// Sleep for 10 seconds to allow talm to reconcile state.
+			// Sometimes if the next test re-creates the IBGUs too quickly,
+			// the policies compliance status is not updated correctly.
+			time.Sleep(10 * time.Second)
+		})
+
 		It("Upgrade end to end", reportxml.ID("68954"), func() {
 			upgradeSuccessful = false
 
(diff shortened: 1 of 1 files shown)


tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
```diff
@@ -100,6 +89,17 @@ var _ = Describe(
...
+
+		if upgradeSuccessful {
+			cnfibuvalidations.PostUpgradeValidations()
+		}
+
+		It("Rollback successful upgrade", reportxml.ID("69058"), func() {
+			if !upgradeSuccessful {
+				Skip("Skipping rollback test due to upgrade failure.")
+			}
+
```
yliu127, line 102: cleanup should not be moved to the end of any test function as this just means the cleanup won't happen when test fails. 

Cleanup step that can potentially affect other test cases, must be added to AfterEach of the test case that creates the resource. 

Sometimes cleanup can be added to the beginning of a test function if the affecting CR may come from external sources that are out of our control.

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
```diff
@@ -50,6 +50,21 @@ var _ = Describe(
 			})
 		})
 
+		AfterEach(func() {
+			By("Deleting IBGUs on target hub cluster", func() {
+				_, err := ibgu.NewIbguBuilder(cnfinittools.TargetHubAPIClient,
+					tsparams.IbguName, tsparams.IbguNamespace).DeleteAndWait(1 * time.Minute)
+
+				Expect(err).ToNot(HaveOccurred(), "Failed to delete prep-upgrade ibgu on target hub cluster")
```
yliu127, line 58: nit: it is no longer "prep-upgrade" ibgu when we clean it up in afterEach. Probably just as simple as Failed to clean up ibgu xxx on target hub cluster.

yliu127 (approved):
/lgtm

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
```diff
@@ -50,6 +50,21 @@ var _ = Describe(
 			})
 		})
 
+		AfterEach(func() {
```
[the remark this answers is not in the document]
yliu127: yes, agreed.

PR #624, opened 2025-05-28
cnf ran: add back in workaround to verify policies

klaskosk:
As a workaround for [OCPBUGS-56702](https://issues.redhat.com/browse/OCPBUGS-56702), this PR adds an extra step in the AfterEach of the oran-post-provision test cases to verify that the policies on the cluster match the ones that should be associated with the ProvisioningRequest. However, the logic has been simplified compared to the old workaround.

Rather than going through the ClusterTemplate to see the expected policies, we wait until the ProvisioningRequest has been fulfilled and check the policy selector label on the ClusterInstance. This will be updated the first time the ProvisioningRequest is fulfilled. From there the process of waiting for the policy version to show up in the policies list is quite similar.

In addition to the extra step in the AfterEach, the same step is added after updating the template version for each test case which does so. This prevents the same issue from occuring in the middle of tests.

### DIFF
--- a/tests/cnf/ran/oran/internal/helper/helper.go
+++ b/tests/cnf/ran/oran/internal/helper/helper.go
@@ -127,3 +127,33 @@ func WaitForValidPRClusterInstance(client *clients.Settings, timeout time.Durati
 			}), nil
 		})
 }
+
+// WaitForPolicyVersion waits up to timeout until all of the policies in the namespace have the specified version.
+// Version is defined as the first hyphen-delimited part of the policy name. Since it lists policies in the spoke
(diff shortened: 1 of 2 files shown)


klaskosk:
Tested 3 times just to be sure this really fixed the issue:
- https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/3289/
- https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/3290/
- https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/3293/

PR #637, opened 2025-06-03
system-tests: fix problem in verify policies are compliant

Demostenes777:
This PR restores the utilization of mutex to avoid problems during the parallel execution of the method VerifyAllPoliciesInNamespaceAreCompliant

@klaskosk if you remember our discussion about this topic. I experienced problems again during parallel execution so I have restored the mutex-based version.


klaskosk (approved):
I'll have to add it to my backlog to figure out the solution to the concurrency issues in eco-goinfra but this workaround lgtm

mcornea (approved):
lgtm

PR #653, opened 2025-06-09
System Test: RDS, back-port cluster log-forwarding test to the 4.18

elenagerman:


elenagerman:
unsupported eco-goinfra method; requires additional back-port for the cluster-logging part for the infra code

PR #658, opened 2025-06-15
cnf network: remove nftables flush all rules from test case

gkopels:
Security nftables test cases remove the nftables stop and flush all rules func.


evgenLevin (approved):
/lgtm

tests/cnf/core/network/security/tests/nftables-custom-firewall-test.go
```diff
@@ -106,9 +106,6 @@ var _ = Describe("nftables", Ordered, Label(tsparams.LabelNftablesTestCases), Co
 		By("Remove the static route to the external Pod network on each worker node")
 		addDeleteStaticRouteOnWorkerNodes(testPodList, routeMap, "del", hubIPv4Network)
 
-		By(fmt.Sprintf("Disables nftables on %s if active", cnfWorkerNodeList[0].Definition.Name))
```
cgoncalves, line 109: This is wrong. This will disable nftables AFTER reverting the custom nftables systemd file, which would then flush ALL rules. Consequently this will cause the issue you are seeing in CI.

Ensure that nftables is disabled before code in lines 118-120:

```
By(fmt.Sprintf("Remove machine-configuration %s", mcNftablesName))
err := mco.NewMCBuilder(APIClient, mcNftablesName).Delete()
Expect(err).ToNot(HaveOccurred(), "Failed to get the machineConfig")
```

tests/cnf/core/network/security/tests/nftables-custom-firewall-test.go
```diff
@@ -502,6 +523,9 @@ func disableNftablesIfActive(nodeName string) {
 			verifyNftablesStatus("inactive", nodeName)
 		}
 	}
+
+	// Reboot the node to ensure clean state
+	rebootNodeAndWaitForMcpStable(nodeName)
```
cgoncalves, line 528: Reboot only when status=='active'

tests/cnf/core/network/security/tests/nftables-custom-firewall-test.go
```diff
@@ -107,11 +107,16 @@ var _ = Describe("nftables", Ordered, Label(tsparams.LabelNftablesTestCases), Co
 		By("Remove the static route to the external Pod network on each worker node")
 		addDeleteStaticRouteOnWorkerNodes(testPodList, routeMap, "del", hubIPv4Network)
 
+		By("Remove nftables entries from machineconfiguration nodeDisruptionPolicy")
+		removeMachineConfigurationNodeDisruptionPolicy()
+
+		By(fmt.Sprintf("Remove machine-configuration %s", mcNftablesName))
+		err := mco.NewMCBuilder(APIClient, mcNftablesName).Delete()
```
cgoncalves, line 114: Deleting the MachineConfig will sooner or later trigger a node reboot, no?
Just wondering if we have to explicitely reboot the node in disableNftablesIfActiveAndReboot or not. Is there a chance that the node gets rebooted twice? The test would still be valid, just an unnecessary reboot cycle. In any case, I'm fine with whatever you choose.

tests/cnf/core/network/security/tests/nftables-custom-firewall-test.go
gkopels: added

PR #659, opened 2025-06-16
system-tests: simplified oran day2 config tests

Demostenes777:
This PR simplifies the verification of the day2 configuration tests. Instead of expecting specific new and old version of PTP and SR-IOV, it only verifies if the PTP version is different after the upgrade. This is sufficient to test day2 configuration upgrades and facilitates CI pipelines maintenance.


Demostenes777:
@klaskosk can we merge this PR?

PR #664, opened 2025-06-18
DRAFT: cnf-network: adapt tests for clusters with Restricted IP forwarding

evgenLevin:
Modified the test logic to support clusters where IP forwarding is restricted by default.

### DIFF
--- a/tests/cnf/core/network/metallb/internal/cmd/cmd.go
+++ b/tests/cnf/core/network/metallb/internal/cmd/cmd.go
@@ -11,7 +11,8 @@ import (
 // DefineRouteAndSleep creates route and appends sleep CMD.
 func DefineRouteAndSleep(dstNet, nextHop string) []string {
 	cmd := DefineRoute(dstNet, nextHop)
-	cmd[2] += " && sleep INF"
+	cmd[2] += ` && (echo 1 > /proc/sys/net/ipv4/ip_forward 2>/dev/null || true); \
+trap : TERM INT; sleep infinity & wait`
 
 	return cmd
 }
--- a/tests/cnf/core/network/metallb/internal/tsparams/mlbvars.go
+++ b/tests/cnf/core/network/metallb/internal/tsparams/mlbvars.go
@@ -85,6 +85,13 @@ var (
 	IBPGPProtocol = "iBGP"
 	// TestLabel represents node label for testing.
 	TestLabel = map[string]string{"test": "label"}
+	// IPForwardAndSleepCmd enables IPv4 forwarding and keep the container running command.
+	IPForwardAndSleepCmd = []string{
+		"/bin/bash",
+		"-c",
+		`echo 1 > /proc/sys/net/ipv4/ip_forward 2>/dev/null || true; \
+trap : TERM INT; sleep infinity & wait`,
+	}
 	// MetalLbBgpMetrics represents the list of expected metallb metrics.
 	MetalLbBgpMetrics = []string{"frrk8s_bgp_keepalives_sent", "frrk8s_bgp_keepalives_received",
 		"frrk8s_bgp_notifications_sent", "frrk8s_bgp_opens_received", "frrk8s_bgp_opens_sent",
--- a/tests/cnf/core/network/metallb/tests/frrk8-tests.go
+++ b/tests/cnf/core/network/metallb/tests/frrk8-tests.go
@@ -405,11 +405,17 @@ var _ = Describe("FRR", Ordered, Label(tsparams.LabelFRRTestCases), ContinueOnFa
 
 				By("Creating FRR Hub pod on worker node 0")
 				_ = createFrrHubPod(hubPodWorker0,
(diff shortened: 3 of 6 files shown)


tests/cnf/core/network/metallb/internal/tsparams/mlbvars.go
```diff
@@ -85,6 +85,13 @@ var (
 	IBPGPProtocol = "iBGP"
 	// TestLabel represents node label for testing.
 	TestLabel = map[string]string{"test": "label"}
+	// IPForwardAndSleepCmd enables IPv4 forwarding and keep the container running command.
+	IPForwardAndSleepCmd = []string{
+		"/bin/bash",
+		"-c",
+		"echo 1 > /proc/sys/net/ipv4/ip_forward 2>/dev/null || true; " +
+			"sleep infinity",
```
karampok, line 93: general note, sleep inf is bad because it does not listen for signals and that has some weird behavior.

Is recommended 
```
/bin/bash -c "trap : TERM INT; sleep infinity & wait"
```

tests/cnf/core/network/metallb/tests/metallb-crds.go
```diff
@@ -97,6 +98,12 @@ var _ = Describe("MetalLb New CRDs", Ordered, Label("newcrds"), ContinueOnFailur
 			WithInterfaces([]string{sriovInterfacesUnderTest[0]}).
 			Create()
 		Expect(err).ToNot(HaveOccurred(), "An unexpected error occurred while creating L2Advertisement.")
+
+		By("Enabling IPForwarding on a DUT interface")
```
karampok, line 102: At the end we will have nmstate var, in between we could have machine config, can we not use the mc instead of doing directly sysctl commands?

tests/cnf/core/network/metallb/tests/frrk8-tests.go
```diff
@@ -479,11 +485,17 @@ var _ = Describe("FRR", Ordered, Label(tsparams.LabelFRRTestCases), ContinueOnFa
 
 				By("Creating FRR Hub pod on worker node 0")
```
karampok, line 486: Just aksing, do we really need the ipforwarding enable  in all pod? I mean unless we have a multihop emulation there should be no forwarding, no?

tests/cnf/core/network/metallb/internal/cmd/cmd.go
```diff
@@ -11,7 +11,8 @@ import (
 // DefineRouteAndSleep creates route and appends sleep CMD.
 func DefineRouteAndSleep(dstNet, nextHop string) []string {
 	cmd := DefineRoute(dstNet, nextHop)
-	cmd[2] += " && sleep INF"
+	cmd[2] += " && echo 1 > /proc/sys/net/ipv4/ip_forward 2>/dev/null 2>/dev/null || true; " +
```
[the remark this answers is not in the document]
evgenLevin: I considered using an init container, but in this case, it adds more complexity than benefits.

Removed the second 2>/dev/null

tests/cnf/core/network/metallb/internal/tsparams/mlbvars.go
```diff
@@ -85,6 +85,13 @@ var (
 	IBPGPProtocol = "iBGP"
 	// TestLabel represents node label for testing.
 	TestLabel = map[string]string{"test": "label"}
+	// IPForwardAndSleepCmd enables IPv4 forwarding and keep the container running command.
+	IPForwardAndSleepCmd = []string{
```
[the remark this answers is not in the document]
evgenLevin: Because this command already includes /bin/bash -c, while in DefineRouteAndSleep, it's added by the DefineRoute function. It could be refactored, but that's not the goal of this PR.

tests/cnf/core/network/metallb/internal/tsparams/mlbvars.go
evgenLevin: added

tests/cnf/core/network/metallb/tests/frrk8-tests.go
evgenLevin: No, we don't. I only enable IP forwarding on hub pods that handle routing between two networks in the multihop scenario.

tests/cnf/core/network/metallb/tests/metallb-crds.go
evgenLevin: Yes, I’ve also considered adding a MachineConfig, but that would trigger node reboots, which significantly increase test runtime — something critical for our new CI setup in PROW.

As an alternative, we could disable the reboot, but that would add complexity to the tests themselves.

karampok: okay, but I hope is ready for nmstate (this is designed not to need a reboot)

karampok (approved):
lgtm
(even if I agree the implementation is not optimal, but is not worth the refactor)

ajaggapa (approved):
lgtm

PR #671, opened 2025-06-24
cnf network: fix nodeselector test and validatePrefix

ajaggapa:
`TestID:53991` uses our `metallb:metallbtests` worker labels which is used across all metallb tests. As part of the test, we are removing the labels which leads to failure of subsequent nodeselector tests failure in a 3 worker node cluster. Instead, updated the test to use different custom label. 

`validatePrefix` checks for `len(bgpStatus.Routes)>0` even when `noRouteCheck is true`. In our tests we are using only single ip pool, so when route check is not required we should have this step. This is a flaky issue, probably passes when the bgp update to remove the route is not received before we are validating.

### DIFF
--- a/tests/cnf/core/network/metallb/tests/common.go
+++ b/tests/cnf/core/network/metallb/tests/common.go
@@ -522,25 +522,35 @@ func validatePrefix(
 ) {
 	var nextHopAddresses []string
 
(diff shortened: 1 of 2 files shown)


evgenLevin (approved):
/lgtm

PR #674, opened 2025-06-25
hwaccel: Nfd testing gc when node are deleted

ggordaniRed:


cdvultur (dismissed):
one note , but overall looks good.

tests/hw-accel/nfd/features/tests/features-test.go
```diff
@@ -240,7 +242,88 @@ var _ = Describe("NFD", Ordered, func() {
...
+			if len(nodes) == 0 {
+				Skip("No worker nodes found to test node deletion")
+			}
+
+			// Select the first worker node for deletion
+			nodeToDelete := nodes[0]
+			glog.V(ts.LogLevel).Infof("Selected node for deletion: %s", nodeToDelete.Object.Name)
+
+			By("Delete the selected node")
+			err = nodeToDelete.Delete()
```
cdvultur, line 271: is the node put back on the system ?
Asking because this is a potential destructive tests which, if there is some other test running, of other operators, might cause problem.
I would suggest adding a flag to run destructive tests, just in case

cdvultur (changes requested):
I would suggest to have this test behind a flag to prevent destructive actions in conjunction with other tests

PR #707, opened 2025-07-18
lca: remove etcd seed hostname references validation test

Demostenes777:
This PR removes a test from LCA that is no longer necessary.


mcornea:
lgtm

PR #711, opened 2025-07-20
cnf ran,system-tests: update oran tests for Metal3

klaskosk:
Depends-on: openshift-kni/eco-goinfra#1077

This PR makes all of the changes required to support the new Metal3 hardware plugin for O-RAN tests. To avoid breaking `go vet` across multiple commits, this single PR includes changes to both the system-tests o-cloud tests and the cnf/ran oran tests. All that has been changed is the names of the Node and NodePool resources. Functionality of these tests remains the same. However, they now must be run using the Metal3 hardware plugin and will no longer support older builds of the oran-o2ims operator.


klaskosk (commented):
I'm a little early to making this PR since we haven't gotten the ocp-olm-setup automation updated (or git configs) updated yet, but once those are ready we should run both the FT and ST O-RAN tests before merging

klaskosk:
FT tests (minus the `oran-provision.go` ones) passed with this PR. For FT the full execution is blocked by [OCPBUGS-59983], but I think it's fine if we merge this regardless, since provisioning does work just not the one scenario

[OCPBUGS-59983]: https://issues.redhat.com/browse/OCPBUGS-59983

yliu127 (dismissed):

PR #722, opened 2025-07-29
hwaccl; First tests to AMD GPU Operator

aopincar:
This update introduces a directory structure and initial tests for the AMD GPU Operator.

The first test involves verifying the NFD (Node Feature Discovery) label. This test ensures that a specific label (“feature.node.kubernetes.io/amd-gpu=true”) was added by NFD to each worker node equipped with AMD GPUs.

The second, and primary, test focuses on one of the key components of the AMD GPU Operator: the Node Labeller. This test begins by enabling the Node Labeller, followed by waiting for the relevant pods to be created and running on all AMD GPU worker nodes. Finally, it validates that all necessary labels have been successfully added.


tests/hw-accel/amdgpu/basic/tests/basic-test.go
```diff
@@ -0,0 +1,242 @@
...
+
+		AfterAll(func() {})
+
+		It("Check AMD label was added by NFD", func() {
+
+			// The assumption is that all Worker Nodes are equipped with AMD GPUs.
+			amdNFDLabelFound, amdNFDLabelFoundErr := get.LabelPresentOnAllNodes(
+				apiClient, tsp.AMDNFDLabelKey, tsp.AMDNFDLabelValue, inittools.GeneralConfig.WorkerLabelMap)
+
+			Expect(amdNFDLabelFoundErr).ToNot(HaveOccurred(),
```
achuzhoy, line 56: it's ToNot(HaveOccurred() here and later I see you use To(BeNil() for error

I'd suggest using the same method for consistency

tests/hw-accel/amdgpu/basic/tests/basic-test.go
```diff
@@ -0,0 +1,242 @@
...
+			nonRunningPodsCnt := len(nodeLabellerPods)
+			for _, nodeLabellerPod := range nodeLabellerPods {
+				for attempt := 1; attempt <= tsp.MaxAttempts; attempt++ {
+					refreshedNodeLabellerPod, refreshedNodeLabellerPodErr := pod.Pull(
+						apiClient, nodeLabellerPod.Object.Name, nodeLabellerPod.Object.Namespace)
+					Expect(refreshedNodeLabellerPodErr).To(BeNil(), fmt.Sprintf(
+						"Failed to pull refreshed Pod name '%v'. Error:\n%v\n",
+						nodeLabellerPod.Object.Name, refreshedNodeLabellerPodErr))
+					glog.V(tsp.AMDGPULogLevel).Infof(
+						"Refreshed Node Labeller Pod '%s' status is '%v'",
```
achuzhoy, line 128: is double 'l' in labeller  correct?

tests/hw-accel/amdgpu/basic/tests/basic-test.go
```diff
@@ -0,0 +1,242 @@
...
+								allNodeLabellerPodsFound = true
+							}
+
+							break
+						}
+					}
+				}
+			}
+
+			Expect(allNodeLabellerPodsFound).To(BeTrue(), "Node Labller Pods haven't been found on all nodes")
```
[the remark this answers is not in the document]
aopincar: Not a typo:
https://github.com/ROCm/k8s-device-plugin/blob/master/cmd/k8s-node-labeller/README.md

tests/hw-accel/amdgpu/basic/tests/basic-test.go
```diff
@@ -0,0 +1,242 @@
...
+
+			Expect(amdNodeBuildersErr).ToNot(HaveOccurred(),
+				fmt.Sprintf("Failed to get Builders for AMD GPU Worker Nodes. Error:\n%v\n", amdNodeBuildersErr))
+
+			Expect(amdNodeBuilders).ToNot(BeEmpty(),
+				"'amdNodeBuilders' can't be empty")
+
+		})
+
+		BeforeEach(func() {})
```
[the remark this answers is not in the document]
aopincar: Done

tests/hw-accel/amdgpu/basic/tests/basic-test.go
aopincar: Done

tests/hw-accel/amdgpu/basic/tests/basic-test.go
```diff
@@ -0,0 +1,236 @@
...
+			// Because a random postfix is added to the name of each Pod, we need to find their name first, and then
+			// to wait for them to reach that status.
+
+			var nodeLabellerPods []*pod.Builder
+			allNodeLabellerPodsFound := false
+			for _, amdNodeBuilder := range amdNodeBuilders {
+				nodeName := amdNodeBuilder.Object.Name
+				podListFieldSelector := fmt.Sprintf("spec.nodeName=%s", nodeName)
+
+				for attempt := 1; attempt <= tsp.MaxAttempts && !allNodeLabellerPodsFound; attempt++ {
```
ggordaniRed, line 84: If you need to use wait and pull take a look at wait.PollUntilContextTimeout from "k8s.io/apimachinery/pkg/util/wait"

tests/hw-accel/amdgpu/basic/tests/basic-test.go
```diff
@@ -0,0 +1,121 @@
...
+			Expect(amdNodeBuildersErr).To(BeNil(),
+				fmt.Sprintf("Failed to get Builders for AMD GPU Worker Nodes. Error:\n%v\n", amdNodeBuildersErr))
+
+			Expect(amdNodeBuilders).ToNot(BeEmpty(),
+				"'amdNodeBuilders' can't be empty")
+
+		})
+
+		It("Check AMD label was added by NFD", func() {
+
```
wabouhamad, line 43: Would help to add a By(.... description of testcase...) step or a glog statement to explain what is tested, so in logs you can see what each testcase is doing

tests/hw-accel/amdgpu/basic/tests/basic-test.go
```diff
@@ -0,0 +1,121 @@
...
+			Expect(amdNFDLabelFoundErr).To(BeNil(),
+				"An error occurred while attempting to verify the AMD label by NFD: %v ", amdNFDLabelFoundErr)
+			Expect(amdNFDLabelFound).To(BeTrue(),
+				"AMD label check failed to match label %s and label value %s on all nodes",
+				amdparams.AMDNFDLabelKey, amdparams.AMDNFDLabelValue)
+		})
+
+		It("Node Labeller", func() {
+
```
wabouhamad, line 56: Same here, for each It() testcase, add a a By(.... description of testcase...) step or a glog statement to explain what is tested, so in logs you can see what each testcase is doing

tests/hw-accel/amdgpu/internal/labels/lables.go
```diff
@@ -0,0 +1,252 @@
...
+	defer cancel()
+
+	errCh := make(chan error, len(labelNodes))
+
+	var waitGroup sync.WaitGroup
+
+	for _, labelNode := range labelNodes {
+		waitGroup.Add(1)
+
+		go LabelsMissingOnNode(ctx, labelNode, labels, timeout, interval, &waitGroup, errCh)
```
ggordaniRed, line 199: In a for loop, Go reuses the same memory location for the loop variable (node in this case) in every single iteration. It just updates the value at that location.

A goroutine often doesn't start executing the instant it's created. The for loop can run to completion before most of the goroutines even get scheduled.

// ❌ Problematic Pattern
for _, node := range labelNodes {
go func() {
// This 'node' variable will be the same for all goroutines
// and will hold the value of the last node.
doWork(node)
}()
}
// ✅ Corrected Pattern
for _, node := range labelNodes {
go func(nodeCopy *nodes.Builder) {
// 'nodeCopy' is now a safe, private copy for this goroutine.
doWork(nodeCopy)<-- in this way you make sure to pass the updated value
}(node) // Pass the current node as an argument here
}

tests/hw-accel/amdgpu/internal/labels/lables.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	waitGroup.Wait()
+
+	if len(errCh) > 0 {
```
ggordaniRed, line 204: probably some of the error messages will be lost. for guarantee a safe read of all error in this channel consider add this after waitGroup.wait() 
close(errCh) // because there are no produces it is safe o close the channel

	// Collect all errors from the channel.
	var allErrors []error
	for err := range errCh {
		allErrors = append(allErrors, err)
	}

	// Use errors.Join to return a single error containing all the others.
	// errors.Join returns nil if the slice is empty.
	return errors.Join(allErrors...)

wabouhamad (approved):
/lgtm

PR #728, opened 2025-07-30
lca ibu: Automating test for states transitions

achuzhoy:
The test verifies it's possible to update states
after prep and also possible to change the seedimage.


trewest (dismissed):
/lgtm

PR #755, opened 2025-08-20
lca ibu: fix kmm failure

achuzhoy:
A field was changed to be optional


achuzhoy:
cc @trewest @cdvultur

PR #808, opened 2025-09-07
 add imagestream validation for preflightvalidationocp tests

ItzikEzra-rh:
- Add HasTag() method to imagestream package to check for specific tags
- Add CheckImageStreamForModule() in get.go to validate imagestream when using internal registry
- Integrate imagestream validation into signing, use-dtk-auto, and realtime tests
- Validation only runs when the module uses the OpenShift internal registry (image-registry.openshift-image-registry.svc)
- Includes polling mechanism (2 min timeout) for imagestream readiness


cdvultur (changes requested):
Some changes are needed.

PR #856, opened 2025-10-05
refactor: Remove inittools dependency from reporter package

bazem8:
- Replace inittools import with direct config import in reporter
- Add local GeneralConfig initialization in reporter package
- Add SetGeneralConfig function for optional configuration override
- Maintain backward compatibility with existing API
- Allow RAN deployment to run without ECO_DRY_RUN=true requirement
- Preserve all existing reporter functionality and behavior

This change enables test suites to use reporter without importing the top-level inittools package, removing the ECO_DRY_RUN=true constraint for RAN deployment tests.

### DIFF
--- a/tests/internal/reporter/reporter.go
+++ b/tests/internal/reporter/reporter.go
@@ -11,13 +11,30 @@ import (
 	"github.com/golang/glog"
 	"github.com/onsi/ginkgo/v2/types"
 	"github.com/openshift-kni/k8sreporter"
-	. "github.com/rh-ecosystem-edge/eco-gotests/tests/internal/inittools"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/internal/config"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
 var (
 	pathToPodExecLogs = "/tmp/pod_exec_logs.log"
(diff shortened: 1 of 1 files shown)


tests/internal/reporter/reporter.go
```diff
@@ -11,14 +11,31 @@ import (
...
 )
 
 var (
 	pathToPodExecLogs = "/tmp/pod_exec_logs.log"
+	// generalCfg holds the configuration for reporter operations
+	generalCfg *config.GeneralConfig
 )
 
+// init initializes the reporter configuration
+func init() {
```
klaskosk, line 25: I think this init is fine given that it's still an improvement. However, we'll need to exclude the `gochecknoinits` from the path `tests/internal/reporter` in the `.golangci.yml` so that the linter passes. Please update that config file to do so

klaskosk (commented):
Overall looks good. Just need to update the linter config and then make some small changes to get the linter passing

klaskosk (approved):

bazem8:
reporting testing:
https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/CI/job/far-edge-vran-deployment/12121/

PR #1006, opened 2025-12-17
Closes CNF-19306; Migration of BC HA test cases from cnf-gotests

dpopsuev:
This is a migration of three test cases from cnf-gotests of the PTP BC HA.
CNF-73093 - 'should change high availability active profile when other nic interface is down'
CNF-73094 - 'should move to FREERUN state when active and inactive interfaces are down'
CNF-73095 - 'should change high availability active profile when active profile is deleted'

## Summary by CodeRabbit

* **New Features**
  * Retrieve HA profile names from metrics for a node.
  * Remove a profile from a configuration with validation and orphaned-recommendation cleanup.

* **Tests**
  * Expanded HA failover coverage: active-profile switch, total HA failure, and profile-deletion scenarios.
  * Added helpers to wait for profile changes, verify HA health, and ensure interface/config restoration.

* **Bug Fixes**
  * Treat missing/restored profiles in configs as benign; preserve original indexes and append restored entries instead of failing.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tests/cnf/ran/ptp/internal/profiles/restoreconfigs.go
+++ b/tests/cnf/ran/ptp/internal/profiles/restoreconfigs.go
@@ -98,8 +98,6 @@ func RestorePtpConfigs(client *clients.Settings, ptpConfigList []*ptp.PtpConfigB
 // listChangedProfilesInConfig lists the profiles in the PtpConfig that have changed. It considers all the profiles in
 // the original PtpConfig and then uses reflect.DeepEqual to compare each to the latest version of the PtpConfig. It
 // returns a list of profile references that were changed.
-//
-// This function will return an error if a profile is not found in the latest PtpConfig.
 func listChangedProfilesInConfig(ptpConfig *ptp.PtpConfigBuilder) ([]*ProfileReference, error) {
 	latestPtpConfig, err := ptpConfig.Get()
 	if err != nil {
@@ -108,28 +106,97 @@ func listChangedProfilesInConfig(ptpConfig *ptp.PtpConfigBuilder) ([]*ProfileRef
 
 	var changedProfiles []*ProfileReference
 
-	for _, profile := range ptpConfig.Definition.Spec.Profile {
+	for profileIndex, profile := range ptpConfig.Definition.Spec.Profile {
 		if profile.Name == nil {
 			return nil, fmt.Errorf("profile name is nil")
 		}
 
-		index := slices.IndexFunc(latestPtpConfig.Spec.Profile, func(p ptpv1.PtpProfile) bool {
+		latestIndex := slices.IndexFunc(latestPtpConfig.Spec.Profile, func(p ptpv1.PtpProfile) bool {
 			return p.Name != nil && *p.Name == *profile.Name
 		})
-		if index == -1 {
-			return nil, fmt.Errorf("failed to find profile %s in latest PtpConfig", *profile.Name)
+
+		if latestIndex == -1 {
+			// restoring profile that was deleted from config
+			changedProfiles = append(changedProfiles, &ProfileReference{
+				ConfigReference: runtimeclient.ObjectKeyFromObject(ptpConfig.Definition),
+				ProfileIndex:    profileIndex,
+				ProfileName:     *profile.Name,
+			})
+
+			continue
 		}
 
-		if reflect.DeepEqual(profile, latestPtpConfig.Spec.Profile[index]) {
+		if reflect.DeepEqual(profile, latestPtpConfig.Spec.Profile[latestIndex]) {
 			continue
 		}
 
 		changedProfiles = append(changedProfiles, &ProfileReference{
 			ConfigReference: runtimeclient.ObjectKeyFromObject(ptpConfig.Definition),
-			ProfileIndex:    index,
+			ProfileIndex:    profileIndex,
 			ProfileName:     *profile.Name,
 		})
 	}
 
 	return changedProfiles, nil
+}
+
+// RemoveProfileFromConfig removes a profile from a PtpConfig by deleting it from the profile array. The
+// profile is identified by the provided ProfileReference.
+func RemoveProfileFromConfig(
+	client *clients.Settings,
+	profileReference ProfileReference,
+) error {
+	ptpConfig, err := profileReference.PullPtpConfig(client)
+	if err != nil {
+		return fmt.Errorf("failed to pull PtpConfig %s: %w",
+			profileReference.ConfigReference.Name, err)
+	}
+
+	profileIndex := profileReference.ProfileIndex
+
+	// Validate index is in bounds
+	if profileIndex < 0 || profileIndex >= len(ptpConfig.Definition.Spec.Profile) {
+		return fmt.Errorf("profile %s at index %d is out of bounds (config has %d profiles)",
+			profileReference.ProfileName, profileIndex, len(ptpConfig.Definition.Spec.Profile))
+	}
+
+	// Verify the profile name matches (safety check)
+	profileToRemove := ptpConfig.Definition.Spec.Profile[profileIndex]
+	actualName := "<nil>"
+
+	if profileToRemove.Name != nil {
+		actualName = *profileToRemove.Name
+	}
+
+	if profileToRemove.Name == nil || actualName != profileReference.ProfileName {
+		return fmt.Errorf("profile name mismatch at index %d: expected %s, got %v",
+			profileIndex, profileReference.ProfileName, actualName)
+	}
+
+	// Remove the profile from the slice
+	ptpConfig.Definition.Spec.Profile = append(
+		ptpConfig.Definition.Spec.Profile[:profileIndex],
+		ptpConfig.Definition.Spec.Profile[profileIndex+1:]...,
+	)
+
+	// Remove orphaned recommend entries
+	filteredRecommend := make([]ptpv1.PtpRecommend, 0)
+
+	for _, rec := range ptpConfig.Definition.Spec.Recommend {
+		// Keep recommendations that don't reference the deleted profile
+		if rec.Profile == nil || *rec.Profile != profileReference.ProfileName {
+			filteredRecommend = append(filteredRecommend, rec)
+		}
+	}
+
+	ptpConfig.Definition.Spec.Recommend = filteredRecommend
+
+	// Update the config in the cluster
+	_, err = ptpConfig.Update()
+	if err != nil {
+		return fmt.Errorf("failed to update PtpConfig after removing profile %s: %w",
+			profileReference.ProfileName, err)
+	}
+
+	return nil
 }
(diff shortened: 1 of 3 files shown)


tests/cnf/ran/ptp/tests/ptp-interfaces.go
```diff
@@ -272,4 +272,465 @@ var _ = Describe("PTP Interfaces", Label(tsparams.LabelInterfaces), func() {
...
+
+			testActuallyRan = true
+
+			startTime := time.Now()
+
+			By(fmt.Sprintf("bringing down the active HA's interface %s", activeInterface))
+			err = iface.SetInterfaceStatus(RANConfig.Spoke1APIClient, nodeName, activeInterface, iface.InterfaceStateDown)
+			Expect(err).ToNot(HaveOccurred(), "Failed to set the active HA's interface down")
+
+			DeferCleanup(func() {
```
klaskosk: minor point, but something Yang caught on one of my PRs that I think is good: we should add the DeferCleanup before we actually bring the interface down just to be safe

I also think it'd be better to restore the interface status and leave the DeferCleanup as a precaution for failures. It runs after all other AfterEach nodes, so it causes issues to use as normal cleanup

tests/cnf/ran/ptp/tests/ptp-interfaces.go
```diff
@@ -272,4 +272,465 @@ var _ = Describe("PTP Interfaces", Label(tsparams.LabelInterfaces), func() {
...
+			for _, inactiveProfile := range inactiveProfiles {
+				inactiveInterface := getHAProfileInterface(nodeInfo, inactiveProfile)
+				inactiveInterfaces = append(inactiveInterfaces, inactiveInterface)
+			}
+
+			By("checking if active interface is the egress interface")
+			egressInterface, err := iface.GetEgressInterfaceName(RANConfig.Spoke1APIClient, nodeName)
+			Expect(err).ToNot(HaveOccurred(), "Failed to get egress interface")
+
+			if activeInterface.GetNIC() == egressInterface.GetNIC() {
```
klaskosk: I may be wrong here, but is GetNIC necessary here? As long as they're not the same exact interface I don't think there's an issue with bringing down a different interface on the same NIC

tests/cnf/ran/ptp/tests/ptp-interfaces.go
```diff
@@ -272,4 +272,465 @@ var _ = Describe("PTP Interfaces", Label(tsparams.LabelInterfaces), func() {
...
+
+			By("validating no HOLDOVER event for original inactive interfaces")
+			eventPod, err := consumer.GetConsumerPodforNode(RANConfig.Spoke1APIClient, nodeName)
+			Expect(err).ToNot(HaveOccurred(), "Failed to get event pod")
+
+			for _, inactiveInterface := range inactiveInterfaces {
+				inactiveNIC := inactiveInterface.GetNIC()
+				holdoverFilter := events.All(
+					events.IsType(eventptp.PtpStateChange),
+					events.HasValue(events.WithSyncState(eventptp.HOLDOVER), events.OnInterface(inactiveNIC)),
```
klaskosk: may be a good improvement to add an `events.OnInterfaces`? although it'd be less convenient than metrics having a single way to say contains for all values...

just thinking out loud, this way is fine for this PR, especially since we're searching the same time range

tests/cnf/ran/ptp/tests/ptp-interfaces.go
```diff
@@ -272,4 +272,465 @@ var _ = Describe("PTP Interfaces", Label(tsparams.LabelInterfaces), func() {
...
+			By("validating CLOCK_REALTIME is in LOCKED state")
+			clockRealtimeQuery := metrics.ClockStateQuery{
+				Interface: metrics.Equals(iface.ClockRealtime),
+				Node:      metrics.Equals(nodeName),
+			}
+			err = metrics.AssertQuery(context.TODO(), prometheusAPI, clockRealtimeQuery, metrics.ClockStateLocked,
+				metrics.AssertWithTimeout(5*time.Minute),
+				metrics.AssertWithStableDuration(10*time.Second))
+			Expect(err).ToNot(HaveOccurred(), "Failed to assert CLOCK_REALTIME is LOCKED")
+
```
klaskosk: please add a place to restore the active interface status and verify lock at the end of the test. With MNO support, I think it's easier than relying on AfterEach and the ordering is better than DeferCleanup

tests/cnf/ran/ptp/tests/ptp-interfaces.go
```diff
@@ -272,4 +272,465 @@ var _ = Describe("PTP Interfaces", Label(tsparams.LabelInterfaces), func() {
...
+			By("getting the active and inactive HA profiles")
+			activeProfiles, err := profiles.GetHAProfiles(context.TODO(), prometheusAPI, nodeName, metrics.HAProfileStatusActive)
+			Expect(err).ToNot(HaveOccurred(), "Failed to get active HA profiles")
+			Expect(len(activeProfiles)).To(Equal(1), "Expected exactly one active HA profile")
+
+			activeProfileInfo := nodeInfo.GetProfileByName(activeProfiles[0])
+			Expect(activeProfileInfo).ToNot(BeNil(), "Failed to find active profile in node info")
+
+			haInterfaces := make([]iface.Name, 0, 2)
```
klaskosk: nit, but the capacity should be 1 if we're only going to be adding 1 interface

klaskosk (commented):
Thanks! I think there's a case to be made to move some of the asynchronous assertions into an ha package under internal, similar to `ptpleap`. Something like `getHAProfileInterface` makes sense (although maybe a sign we need more helpers in the profiles package?), but I'm still not settled on the issue

I focused mainly on the first case since there's similarities, erring on the side of more particular, but feel free to disregard if it's still in progress. The main point, which is pretty implicit to be fair, is that I'd like to keep `DeferEach` just for "backup" cleanup and do the main cleanup in the test cases

Overall though, I think this is pretty good and highlights some opportunities for expanding our helpers an finding ways to cut down on boilerplate. I wanted to let the suite evolve a bit before doing too much to address it, so worth keeping in mind solutions

tests/cnf/ran/ptp/tests/ptp-interfaces.go
dpopsuev: @klaskosk I've thought about a pattern I would like to try :D but I must remain focused! I've wrote it down & I'll share it once we are done with the migration. :)

tests/cnf/ran/ptp/tests/ptp-interfaces.go
dpopsuev: It's supposed to be 2, I didn't use the slice correctly, it supposed to be for all interfaces.

tests/cnf/ran/ptp/tests/ptp-interfaces.go
dpopsuev: Updated!

tests/cnf/ran/ptp/internal/profiles/restoreconfigs.go
```diff
@@ -117,7 +122,19 @@ func listChangedProfilesInConfig(ptpConfig *ptp.PtpConfigBuilder) ([]*ProfileRef
 			return p.Name != nil && *p.Name == *profile.Name
 		})
 		if index == -1 {
-			return nil, fmt.Errorf("failed to find profile %s in latest PtpConfig", *profile.Name)
+			// restoring profile that was deleted from config
+			changedProfiles = append(changedProfiles, &ProfileReference{
+				ConfigReference: runtimeclient.ObjectKeyFromObject(ptpConfig.Definition),
+				ProfileIndex:    len(latestPtpConfig.Spec.Profile), // Append at end
```
klaskosk: I regret keeping the changed profiles a little since it makes it more complicated here when we really only check did any change or did none change. It's probably a lot more work than necessary to go back in this PR though

regardless, the profileIndex should be the index in ptpConfig since it already exists there and that's the config we want to go back to

tests/cnf/ran/ptp/internal/profiles/restoreconfigs.go
```diff
@@ -133,3 +150,53 @@ func listChangedProfilesInConfig(ptpConfig *ptp.PtpConfigBuilder) ([]*ProfileRef
...
+		return fmt.Errorf("profile %s at index %d is out of bounds (config has %d profiles)",
+			profileReference.ProfileName, profileIndex, len(ptpConfig.Definition.Spec.Profile))
+	}
+
+	// Verify the profile name matches (safety check)
+	profileToRemove := ptpConfig.Definition.Spec.Profile[profileIndex]
+	if profileToRemove.Name == nil || *profileToRemove.Name != profileReference.ProfileName {
+		return fmt.Errorf("profile name mismatch at index %d: expected %s, got %v",
+			profileIndex, profileReference.ProfileName,
+			func() string {
```
klaskosk: out of curiosity, did gpt generate this function? I've noticed it do inline functions like this but not in other models so I'm curious if it's one of its quirks

a nit though is that it'd be more idiomatic to use a separate variable instead of the inline function like this:
```go
		name := "<nil>"
		if profileToRemove.Name != nil {
			name = *profileToRemove.Name
		}
		
		return fmt.Errorf("profile name mismatch at index %d: expected %s, got %s",
			profileIndex, profileReference.ProfileName, name)
```

klaskosk (commented):
starting to regret some of my restoreconfigs design decisions 🙃 

but tests look good, thanks!

tests/cnf/ran/ptp/tests/ptp-interfaces.go
```diff
@@ -272,4 +273,494 @@ var _ = Describe("PTP Interfaces", Label(tsparams.LabelInterfaces), func() {
...
+
+			return false, err
+		}
+
+		// Must have exactly one active profile
+		if len(activeProfiles) != 1 {
+			klog.V(tsparams.LogLevel).Infof("Expected 1 active profile, got %d", len(activeProfiles))
+
+			return false, fmt.Errorf("expected 1 active profile, got %d", len(activeProfiles))
+		}
```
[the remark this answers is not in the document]
dpopsuev: Keeping it as is for now, if this is becoming a source of flake instead of a source of validation we can fix this later.

tests/cnf/ran/ptp/internal/profiles/restoreconfigs.go
dpopsuev: Working with Cluade-4.5-sonnet.
Good point, I'll look to improve readability.

tests/cnf/ran/ptp/internal/profiles/restoreconfigs.go
klaskosk: it's fine for now :slightly_smiling_face:. since you changed the ProfileIndex the current code is at least correct which is what really counts

klaskosk (approved):
lgtm! Gemini said I should approve it too 😄

PR #1165, opened 2026-02-10
 cnf ran ptp: make OC cases events-optional 

klaskosk:
This commit uses the eventmetric package to fallback to metric
assertions whenever PTP events are not enabled.

Additionally, it flips the semantics of CurrentState. By default, the
CurrentState logs are excluded since they don't represent events being
sent to the consumer. They may be explicitly enabled when desired.

Closes: CNF-21255
Depends-on: #1123 
Assisted-by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Combined event+metric assertion framework for PTP state validation (node/interface, start time, timeout).
  * Runtime check to skip event operations when cluster event publishing is disabled.
  * Tests temporarily tighten Prometheus scrape interval during runs and restore it afterwards.

* **Bug Fixes**
  * Improved error handling and clearer messages around event/config/version checks.

* **Refactor**
  * Tests reworked to use metric-driven assertions for LOCKED, FREERUN, HOLDOVER flows.

* **Documentation**
  * Event option renamed to control inclusion of current-state events (default: exclude).

* **Breaking Changes**
  * Event-wait API/option semantics changed; callers must adopt the new assertion-based flow or updated options.

### DIFF
--- a/tests/cnf/ran/internal/ranparam/const.go
+++ b/tests/cnf/ran/internal/ranparam/const.go
@@ -45,6 +45,8 @@ const (
 	PtpOperatorNamespace = "openshift-ptp"
 	// LinuxPtpDaemonsetName is the name of the Linux PTP daemon daemonset.
 	LinuxPtpDaemonsetName = "linuxptp-daemon"
+	// PtpServiceMonitorName is the name of the PTP operator's ServiceMonitor.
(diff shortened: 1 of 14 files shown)


klaskosk:
this work is now obsolete, any continuation of the concept will be handled in a follow up PR

PR #1322, opened 2026-04-09
Switch ran eco gotests version comparison to semver CNF-20904

bazem8:
switch RAN eco-gotests version comparison to semver - CNF-20904
passed the review of "/cnf-ran-review" . 

## Summary by CodeRabbit

* **Bug Fixes**
  * Version gating now uses full semantic-version semantics with exclusive upper bounds for more accurate compatibility checks.
  * Normalized minimum-version strings across checks (e.g., `4.12.0-0`) and adjusted several gating thresholds (some lowered to `4.11`, others updated to `4.17`/`4.18`/`4.21` variants).
  * Tuned test timing: some wait timeouts reduced for faster feedback; one metric check timeout increased for stability.

* **Chores**
  * Promoted the semantic-version library to a direct dependency.


tests/cnf/ran/internal/version/semver_range.go
```diff
@@ -0,0 +1,148 @@
+package version
```
klaskosk, line 1: minor point, but since we're just leaving a single function in the old file, we can just move all this content to the old file instead of having a separate new file

tests/cnf/ran/internal/version/semver_range.go
```diff
@@ -0,0 +1,148 @@
...
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/Masterminds/semver/v3"
+)
+
+// IsVersionStringInRange reports whether version satisfies minimum <= version < maximumUpper using SemVer 2.0
```
klaskosk, line 11: I think we can simplify the maximum handling by making it exclusive. Similar to how slicing an array like `arr[1:3]` means grab elements `arr[1]` and `arr[2]` but **not** `arr[3]`, we could do the same thing with versions. The math term would be a "half-open interval"

The reason this is nicer is that there's a clear minimum for a given y-stream (`x.y.0-0`), but there's no clear maximum

So if we wanted to do all z-streams for 4.16 and 4.17, we could write the function call as `IsVersionStringInRange(myVersionString, "4.16.0-0", "4.18.0-0")` and then make sure the parsed myVersion is greater than or equal to (equivalently, not less than) `4.16.0-0` and less than `4.18.0-0`

go.mod
```diff
@@ -292,8 +292,8 @@ require (
 )
 
 require (
-	github.com/rh-ecosystem-edge/eco-goinfra v0.0.0-20260408134752-5af19f153594
-	k8s.io/apiextensions-apiserver v0.34.5
+	github.com/Masterminds/semver/v3 v3.4.0
+	github.com/rh-ecosystem-edge/eco-goinfra v0.0.0-20260324223608-9209edd329fe
 )
```
klaskosk, line 297: nit: you can add the `github.com/Masterminds/semver/v3 v3.4.0` line to the top `require` block

klaskosk (commented):
I suspect it'll be hard to avoid merge conflicts with a PR this big, but those also need fixed before merging. 1 main comments about how we handle maximums, I think it'll simplify the process there. Otherwise just minor things

tests/cnf/ran/internal/version/version.go
```diff
@@ -171,25 +120,148 @@ func GetZTPSiteGenerateImage(client *clients.Settings) (string, error) {
...
+// If version is not a valid semver string and maximum is empty, the function returns (true, nil) for compatibility
+// with legacy call sites (e.g. empty or non-semver operator tags with no upper bound). Callers that require a parsed
+// version should validate the string separately or pass a non-empty maximum so the result is (false, nil).
+func IsVersionStringInRange(version, minimum, maximum string) (bool, error) {
+	minV, err := parseMinimumBound(minimum)
+	if err != nil {
+		return false, err
+	}
+
+	maxExclusive, err := parseMaximumExclusiveUpper(maximum)
```
klaskosk: I think we can just change the 3 ish places that use the maximum argument rather than keeping the complex parsing around. That was the point of the previous comment, we can follow the same parsing logic as the minimum and just use a half open interval. That way we can enforce the version strings are semver rather than `X.Y` or other nonstandard formats

klaskosk (commented):
I think there was a bit of a misunderstanding about my comment on the maximum, let me know if want me to clarify some more

also a good thing to do (and tell AI to do) is run `make lint` before committing

tests/cnf/ran/talm/tests/talm-precache.go
```diff
@@ -626,7 +626,7 @@ func assertPrecacheStatus(spokeName, expected string) {
 			cguBuilder.Object.Name, spokeName, cguBuilder.Object.Status.Precaching.Status[spokeName])
 
 		return cguBuilder.Object.Status.Precaching.Status[spokeName]
-	}, 30*time.Minute, 15*time.Second).Should(Equal(expected))
+	}, 20*time.Minute, 15*time.Second).Should(Equal(expected))
```
klaskosk: this line shouldn't be changed by the PR, it should stay 30

tests/cnf/ran/internal/version/version.go
```diff
@@ -171,25 +120,118 @@ func GetZTPSiteGenerateImage(client *clients.Settings) (string, error) {
...
+
+// parseMaximumExclusiveUpper returns the first version not allowed: valid versions satisfy v < upperExclusive.
+// maximum uses the same coercion and parsing rules as minimum; pass a semver-compatible exclusive upper.
+func parseMaximumExclusiveUpper(maximum string) (*semver.Version, error) {
+	if maximum == "" {
+		return nil, nil //nolint:nilnil // empty maximum means no upper bound
+	}
+
+	return parseSemverGateBound(maximum, "maximum")
+}
```
klaskosk: ~~I'd prefer we just use the `parseSemverGateBound` function directly since it's clear it's the same logic~~

actually if we could just remove these altogether based on the other comment that'd be even better since it's simpler

tests/cnf/ran/internal/version/version.go
```diff
@@ -171,25 +120,118 @@ func GetZTPSiteGenerateImage(client *clients.Settings) (string, error) {
...
+
+// parseSemverGateBound trims a leading v, coerces raw to a semver string, and parses it.
+// boundLabel is "minimum" or "maximum" for error messages.
+func parseSemverGateBound(raw string, boundLabel string) (*semver.Version, error) {
+	s := trimSemverVPrefix(raw)
+
+	coerced, err := coerceSemverCore(s)
+	if err != nil {
+		return nil, fmt.Errorf("invalid %s provided: '%s'", boundLabel, raw)
+	}
```
klaskosk: Seeing this now that we've simplified the maximum logic, I don't think that coerceSemverCore is necessary anymore

or actually, that we even need this function at all... if we're just doing `parsedVersion, err := semver.NewVersion(trimSemverVPrefix(version))` in the function, I think it's way simpler to apply the same logic to the bounds as well. That way the only helper function we need is `trimSemverVPrefix`, which is pretty lightweight

klaskosk (commented):
I think there's a couple ways we can simplify further still. My goal is to shift most of the parsing into the purpose-built semver library, so the less we have to do in our package, the better

1 small comment too about a line that shouldn't be changed. But the updates to all the function calls look good now

klaskosk (approved):
lgtm! passes the cnf-ran-review skill too!

PR #1340, opened 2026-04-21
system-tests: Use bootid to check reboot

laurayang842:
Updated hard-reboot.go and nmi-redfish.go to address the issue that master nodes reboot fast enough that the NotReady state is not observed. The issue is often observed on Dell R660 servers. The tests now check for a boot ID change as proof of reboot.

## Summary by CodeRabbit

* **Tests**
  * Enhanced reboot and reset verification in system tests to use bootID-based detection for more reliable node state validation.


yprokule (approved):
/lgtm

PR #1343, opened 2026-04-21
system-tests: update ipvlan-validation.go to support IPv6 single stack

laurayang842:
## Summary by CodeRabbit

* **Tests**
  * Improved IPVLAN connectivity tests to iterate over configured IPv4/IPv6 targets instead of hard-coded permutations.
  * Enforces that at least one target address is configured and skips empty targets.
  * Preserves source/destination directionality across checks.
  * Adds clearer per-deploy expectation messages and enhanced logging for diagnostics.

### DIFF
--- a/tests/system-tests/rdscore/internal/rdscorecommon/ipvlan-validation.go
+++ b/tests/system-tests/rdscore/internal/rdscorecommon/ipvlan-validation.go
@@ -158,33 +158,53 @@ func VerifyIPVlanOnDifferentNodes() {
 // VerifyIPVLANConnectivityBetweenDifferentNodes verifies connectivity between workloads,
 // using IPVLAN interfaces and running on different nodes.
 func VerifyIPVLANConnectivityBetweenDifferentNodes() {
-	verifySRIOVConnectivity(
-		RDSCoreConfig.IPVlanNSOne,
-		RDSCoreConfig.IPVlanNSOne,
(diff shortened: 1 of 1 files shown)


tests/system-tests/rdscore/internal/rdscorecommon/ipvlan-validation.go
```diff
@@ -155,36 +155,40 @@ func VerifyIPVlanOnDifferentNodes() {
 	VerifyIPVLANConnectivityBetweenDifferentNodes()
 }
 
+// verifyIPVLANTargets runs verifySRIOVConnectivity between srcPodLabel and dstPodLabel in namespace
+// for each non-empty address.
+func verifyIPVLANTargets(namespace, srcPodLabel, dstPodLabel, ipv4Target, ipv6Target, expectMsg string) {
```
yprokule: I suggest to have 2 params for namespaces as they could differ(they don't now but that's a valid case)

yprokule (changes requested):

tests/system-tests/rdscore/internal/rdscorecommon/ipvlan-validation.go
laurayang842: updated. Thanks for the comment.

yprokule (approved):
/lgtm

PR #1354, opened 2026-04-28
ocp-sriov: use VFNum instead of hardcoded values in SR-IOV tests

zhiqiangf:
## Problem

Three SR-IOV test files hardcode the numVfs argument in \`NewPolicyBuilder\` calls instead of using \`SriovOcpConfig.VFNum\` (\`ECO_OCP_SRIOV_VF_NUM\`):

- \`exposemtu.go\`: hardcodes \`5\` in 3 places
- \`reinstallation.go\`: hardcodes \`5\` in 3 places
- \`metricsExporter.go\`: hardcodes \`6\` in 3 places

On Mellanox CX6-DX hardware, the Mellanox SR-IOV plugin uses exact-match semantics for the firmware \`NUM_OF_VFS\` field: if the requested VF count differs from the firmware value, it writes the new value to firmware and triggers a node reboot. This reboot takes longer than \`MCOWaitTimeout\` (35 minutes) on this hardware.

**Effect:** When test groups run sequentially with different hardcoded VF counts, each transition triggers a firmware write and reboot that exceeds the timeout, causing tests to fail with \`context deadline exceeded\`.

## Fix

Replace all hardcoded VF counts with \`SriovOcpConfig.VFNum\` in all three files (9 total substitutions). This ensures every test group requests the same VF count, eliminating firmware changes between test groups and the resulting reboot loop.

## Files changed

- \`tests/ocp/sriov/tests/exposemtu.go\`: 3 occurrences of \`5\` → \`SriovOcpConfig.VFNum\`
- \`tests/ocp/sriov/tests/reinstallation.go\`: 3 occurrences of \`5\` → \`SriovOcpConfig.VFNum\`
- \`tests/ocp/sriov/tests/metricsExporter.go\`: 3 occurrences of \`6\` → \`SriovOcpConfig.VFNum\`

## Tests affected

Fixes: 73786, 73787, 73788, 73789, 73790 (exposemtu)
Also fixes the same class of failure for reinstallation tests (46528, 46532) and metricsExporter tests on Mellanox hardware.

### DIFF
--- a/tests/ocp/sriov/tests/exposemtu.go
+++ b/tests/ocp/sriov/tests/exposemtu.go
@@ -93,7 +93,7 @@ var _ = Describe("SRIOV: Expose MTU:", Ordered, Label(tsparams.LabelExposeMTUTes
 			sriovAndResourceName5000,
 			SriovOcpConfig.SriovOperatorNamespace,
 			sriovAndResourceName5000,
-			5,
+			SriovOcpConfig.VFNum,
 			[]string{fmt.Sprintf("%s#0-1", sriovInterfacesUnderTest[0])}, SriovOcpConfig.WorkerLabelMap).
 			WithDevType("netdevice").WithMTU(5000).Create()
 		Expect(err).ToNot(HaveOccurred(), "Failed to configure SR-IOV policy with mtu 5000")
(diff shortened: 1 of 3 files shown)


evgenLevin (approved):
lgtm

PR #1355, opened 2026-04-29
ocp-sriov: fix spoof check verification to use VF index instead of MAC

zhiqiangf:
## Problem

\`verifySpoofCheck\` identifies the pod's VF in \`ip link show <PF>\` by looking for a line containing the pod's static MAC address. On Mellanox CX6-DX hardware, the SR-IOV CNI sets the MAC only inside the container's network namespace (via \`ip link set net1 address <mac>\`) without propagating it to the host PF via \`ip link set <PF> vf N mac\`. As a result, the host-side VF line shows \`00:00:00:00:00:00\` and the MAC-based lookup never finds the expected spoof check state, causing tests 25959 and 70820 to fail consistently on this hardware.

## Fix

Use the VF PCI address from the pod's \`k8s.v1.cni.cncf.io/network-status\` annotation to find the VF's index by matching against the PF's \`virtfnN\` sysfs symlinks. Then look up that specific \`vf N\` line in \`ip link show <PF>\` to verify the spoof check state. This approach is hardware-agnostic and independent of whether the CNI propagates the MAC to the host PF.

The MAC-based path is kept as a fallback for environments where the PCI address is not available in the network-status annotation.

## Changes

- Refactored \`verifySpoofCheck\` into three functions:
  - \`verifySpoofCheck\` — entry point, selects PCI or MAC strategy
  - \`verifySpoofCheckByPCI\` — primary: resolves VF index from PCI address via \`virtfnN\` symlinks, checks specific VF line
  - \`verifySpoofCheckByMAC\` — fallback: existing MAC-matching behavior
- Extracted \`nodeOutput\` helper to deduplicate FQDN/short-hostname lookup logic

## Tests fixed

- 25959: SR-IOV VF with spoof checking enabled
- 70820: SR-IOV VF with spoof checking disabled

## Summary by CodeRabbit

* **Tests**
  * Refactored spoof-check verification to prefer PCI-address-based validation when available, with a fallback to MAC-based checks.
  * Removed the previous MAC-only verification path for more authoritative results.
  * Added a utility to more reliably select per-node command output, improving test robustness and consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/eco-gotests/pull/1355)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -64,6 +64,7 @@ linters:
             - github.com/openshift
             - github.com/nmstate/kubernetes-nmstate
             - github.com/hashicorp/go-version
+            - github.com/Masterminds/semver/v3
             - github.com/cavaliergopher/grab/v3
             - github.com/k8snetworkplumbingwg
             - github.com/metallb/metallb-operator
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ toolchain go1.26.0
 require (
 	github.com/BurntSushi/toml v1.6.0
 	github.com/Juniper/go-netconf v0.3.1
+	github.com/Masterminds/semver/v3 v3.4.0
 	github.com/Masterminds/sprig/v3 v3.3.0
 	github.com/NVIDIA/gpu-operator v1.11.1
 	github.com/cavaliergopher/cpio v1.0.1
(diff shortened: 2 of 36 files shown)


evgenLevin (approved):
/lgtm

PR #1357, opened 2026-04-29
lca ipc: update the dns validation

achuzhoy:
This change follows the upstream Kubernetes apimachinery update that now passes the real value to field.Invalid(), resulting in error messages like Invalid value: "192.168.130.261" instead of Invalid value: "string".

## Summary by CodeRabbit

* **Tests**
  * Refined DNS server configuration validation error messages to include the specific invalid IP address value, providing more informative feedback when misconfiguration occurs.


achuzhoy:
cc @trewest

trewest (approved):

PR #1362, opened 2026-05-01
lca ipc: support different check for dnsServers

achuzhoy:
The test is different for versions below 4.22
Adding support for both variants.


achuzhoy:
cc @trewest

PR #1368, opened 2026-05-03
kmm modules: fix nil pointer panic in build-without-loading test

ItzikEzra-rh:
## Summary
- Delete stale configmap before creating it in the "should build image without loading it" test
- The configmap `simple-kmod` is created by Context 1 and never cleaned up before Context 2 reuses the same name
- The common framework's `Create()` returns nil error on `AlreadyExists` but does not set `builder.Object`, causing a nil pointer dereference at `dockerfileConfigMap.Object.Name`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Enhanced test robustness by adding cleanup logic to ensure consistent state initialization before test execution.


cdvultur (approved):
/lgtm

PR #1381, opened 2026-05-07
CNF-20475: migrate tbc ttsc holdover tests dpopsuev

dpopsuev:
Closes https://redhat.atlassian.net/browse/CNF-20475

## Summary by CodeRabbit

* **New Features**
  * Added T-BC Transmitter profile type and improved profile parsing to better detect T-BC/T-TSC roles
  * Added holdover plugin settings management and automatic upstream-port detection for profiles
  * Added holdover label for PTP metric identification
  * Extended PTP metrics with additional clock class values and a new process type

* **Bug Fixes**
  * Fixed consumer service address formatting to improve connectivity

* **Tests**
  * Added end-to-end holdover behavior tests for T-BC and T-TSC (gated for operator 4.20+)

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/eco-gotests/pull/1381)

### DIFF
--- a/tests/cnf/ran/ptp/internal/profiles/parser.go
+++ b/tests/cnf/ran/ptp/internal/profiles/parser.go
@@ -14,7 +14,11 @@ type configSections = map[string]map[string]string
 
 // parsePtpProfile parses the PTP profile and the ptp4l information to get the interfaces and their types before making
 // a determination on the profile type. Maps in the parsedPtp4lConf struct are guaranteed to not be nil when returned.
-func parsePtpProfile(profile ptpv1.PtpProfile, reference ProfileReference) (*ProfileInfo, error) {
+func parsePtpProfile(
+	profile ptpv1.PtpProfile,
+	reference ProfileReference,
+	controlledNames map[string]bool,
+) (*ProfileInfo, error) {
 	profileInfo := &ProfileInfo{
 		Reference: reference,
 	}
@@ -49,7 +53,7 @@ func parsePtpProfile(profile ptpv1.PtpProfile, reference ProfileReference) (*Pro
 		interfaceInfo.Profile = profileInfo
 	}
 
-	profileInfo.ProfileType, err = determineProfileType(profileInfo.Interfaces, profile)
+	profileInfo.ProfileType, err = determineProfileType(profileInfo.Interfaces, profile, controlledNames)
 	if err != nil {
 		return nil, fmt.Errorf("failed to determine profile type: %w", err)
 	}
@@ -146,10 +150,14 @@ func getInterfacesFromPtp4lSections(clientFlag bool, sections configSections) ma
 	return interfaces
 }
 
-// determineProfileType determines the PTP profile type based on the number of interfaces and their clock types.
-// Additionally, it makes use of ts2phc settings to determine if the profile is GM or MultiNICGM. An error is returned
-// if the profile type cannot be determined.
-func determineProfileType(interfaces map[iface.Name]*InterfaceInfo, profile ptpv1.PtpProfile) (PtpProfileType, error) {
+// determineProfileType determines the PTP profile type based on the number of interfaces, their clock types, and
+// cross-profile context. The controlledNames map contains profile names referenced by a TBC transmitter's
+// controllingProfile setting, enabling distinction between T-BC receivers and standalone T-TSC/OC profiles.
+func determineProfileType(
+	interfaces map[iface.Name]*InterfaceInfo,
+	profile ptpv1.PtpProfile,
+	controlledNames map[string]bool,
+) (PtpProfileType, error) {
 	// If the profile has chronyd configuration, it is a NTP fallback profile. This must be checked before the GM
 	// profile is determined, otherwise the profile would be incorrectly identified as a GM profile.
 	if profile.ChronydConf != nil && *profile.ChronydConf != "" {
@@ -186,7 +194,18 @@ func determineProfileType(interfaces map[iface.Name]*InterfaceInfo, profile ptpv
 		}
 	}
 
+	profileName := ""
+	if profile.Name != nil {
+		profileName = *profile.Name
+	}
+
 	switch {
+	// If the profile has one client interface and is referenced by a TBC transmitter, return ProfileTypeTBCReceiver.
+	case numInterfaces == 1 && numClientInterfaces == 1 && controlledNames[profileName]:
+		return ProfileTypeTBCReceiver, nil
+	// If the profile has one client interface with ts2phc and phc2sys (telecom slave), return ProfileTypeTTSC.
+	case numInterfaces == 1 && numClientInterfaces == 1 && hasTelecomSlaveConfig(profile):
+		return ProfileTypeTTSC, nil
 	// If the profile has one interface and one client interface, return ProfileTypeOC.
 	case numInterfaces == 1 && numClientInterfaces == 1:
 		return ProfileTypeOC, nil
@@ -196,6 +215,9 @@ func determineProfileType(interfaces map[iface.Name]*InterfaceInfo, profile ptpv
 	// If the profile has at least two interfaces and only one client interface, return ProfileTypeBC.
 	case numInterfaces >= 2 && numClientInterfaces == 1:
 		return ProfileTypeBC, nil
+	// If all interfaces are server, return ProfileTypeTBCTransmitter.
+	case numInterfaces >= 1 && numServerInterfaces == numInterfaces:
+		return ProfileTypeTBCTransmitter, nil
 	// All other profile types are considered unsupported.
 	default:
 		return 0, fmt.Errorf("unable to determine PTP profile type based on defined rules")
@@ -235,4 +257,19 @@ func hasClientFlag(ptp4lOpts *string) bool {
 	}
 
 	return false
+}
+
+// hasTelecomSlaveConfig checks whether a profile has the configuration markers of a Telecom Time Slave Clock
+// (ITU-T G.8275.1): ts2phc for hardware timestamping and phc2sys for system clock synchronization, without
+// being a ts2phc master (which would indicate a GM profile).
+func hasTelecomSlaveConfig(profile ptpv1.PtpProfile) bool {
+	if profile.Ts2PhcConf == nil || profile.Phc2sysOpts == nil {
+		return false
+	}
+
+	if strings.Contains(*profile.Ts2PhcConf, "ts2phc.master 1") {
+		return false
+	}
+
+	return true
 }
--- a/tests/cnf/ran/ptp/internal/profiles/plugins.go
+++ b/tests/cnf/ran/ptp/internal/profiles/plugins.go
@@ -8,6 +8,7 @@ import (
 	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/ptp"
 	ptpv1 "github.com/rh-ecosystem-edge/eco-goinfra/pkg/schemes/ptp/v1"
 	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/iface"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 )
 
 // PinStateType enumerates the supported pin states.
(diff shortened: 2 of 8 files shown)


tests/cnf/ran/ptp/internal/tsparams/consts.go
```diff
@@ -25,6 +25,8 @@ const (
 	LabelStability = "stability"
 	// LabelLogReduction is the label for all tests in the PTP log reduction suite.
 	LabelLogReduction = "log-reduction"
+	// LabelHoldover is the label for all tests in the PTP holdover suite.
+	LabelHoldover = "holdover"
```
klaskosk: there's a few different T-GM holdover tests that are probably better in a separate file/section of the suite, so maybe we should be more specific than `holdover`

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,1092 @@
...
+			controlled[name] = true
+		}
+	}
+
+	return controlled
+}
+
+// isTimeSlaveClockProfile checks whether a PTP profile is configured as a time slave clock (T-TSC). A TSC must
+// have phc2sys and ts2phc configurations, must not be a ts2phc master, and must not have masterOnly interfaces.
+func isTimeSlaveClockProfile(profile ptpv1.PtpProfile) bool {
```
klaskosk: ideally we reuse the same profile type classification for the TSC profiles (although looking at these criteria it seems like an OC profile would meet all of them)

tests/cnf/ran/ptp/internal/profiles/parser.go
```diff
@@ -196,6 +196,9 @@ func determineProfileType(interfaces map[iface.Name]*InterfaceInfo, profile ptpv
 	// If the profile has at least two interfaces and only one client interface, return ProfileTypeBC.
 	case numInterfaces >= 2 && numClientInterfaces == 1:
 		return ProfileTypeBC, nil
+	// If the profile has at least two interfaces and all are server, return ProfileTypeTBCTransmitter.
+	case numInterfaces >= 2 && numServerInterfaces > 0 && numClientInterfaces == 0:
```
klaskosk: correct me if not, but I don't think there's a reason there has to be 2 or more server interfaces. In that case, it'd be more accurate to say `numInterfaces >= 1 && numServerInterfaces == numInterfaces`

tests/cnf/ran/ptp/internal/profiles/parser.go
klaskosk: ideally we'd be able to classify a `ProfileTypeTBCReceiver` and `ProfileTypeTTSC` in this function. From what I understand, the main difficulty there is that these profiles are identical. It ends up becoming context-sensitive in that whether a profile is T-BC TR or T-TSC depends on whether another profile which has controllingProfile set to the current profile

this should be achievable with the current parser, although it'll be a bit ugly. I think we'd need to compute the `controlledNames` (`controllingNames` may be more accurate) in `GetNodeInfoMap` per node before iterating through all `recommendsForNode`

it's made worse by my now-regrettable decision to have `getMemoizedAndClonedProfileInfo` handle the matching of profile reference to profiles. it may just be easier to do a slight refactor so that a map of profile references to profiles is calculated for a node, then we collect `controllingNames` based on that map, then we call `getMemoizedAndClonedProfileInfo` with a single reference + profile and the `controllingNames` map

the plan is to overhaul the parser and `GetNodeInfoMap` but I'd prefer to finish up the migration first

---

tl;dr I made some poor design decisions in the parser but it should be possible to classify T-BC TR and T-TSC within the `determineProfileType` function

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,1092 @@
...
+			return name, nil
+		}
+	}
+
+	return pluginName, nil
+}
+
+// interconnectionsUpstreamPort extracts the upstreamPort value from the plugin's interconnections section
+// for the specified profile index.
+func interconnectionsUpstreamPort(ptpConfig *ptpv1.PtpConfig, profileIndex int) (string, error) {
```
klaskosk: could we just use the port mentioned in the config? it seems like it's necessarily the same as the upstream port for T-BC/T-TSC TR profiles. it'd also be a lot easier to check since we're already parsing the ports in the ptp4lConf

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,1092 @@
...
+		}
+
+		return portStr, nil
+	}
+
+	return "", fmt.Errorf("key 'upstreamPort' not found in interconnections of plugin")
+}
+
+// getPluginSettings reads the holdover plugin settings from the specified profile in the PtpConfig.
+func getPluginSettings(ptpConfig *ptpv1.PtpConfig, profileIndex int) (*holdoverPluginSettings, error) {
```
klaskosk: `profiles/plugins.go` already contains a lot of helpers related to the plugins. you could create a new function which reuses `getIntelPlugin` to return the holdover settings (and define `HoldoverPluginSettings` in that file)

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,1092 @@
...
+	if profile.Ptp4lConf != nil && strings.Contains(*profile.Ptp4lConf, "masterOnly 1") {
+		return false
+	}
+
+	return true
+}
+
+// getPluginName extracts the plugin name from a PTP profile. Returns the first plugin name with settings if
+// found, otherwise returns the last plugin name. Returns an error if no plugins are configured.
+func getPluginName(profile *ptpv1.PtpProfile) (string, error) {
```
klaskosk: `profiles/plugins.go` already defines `GetPluginTypesFromProfile`. I think it's worth breaking down the use cases for knowing the plugin name: excluding GNR-D, getting upstream port, and getting/setting plugin settings. it may be easier to just explicitly use `PluginTypeE810` and then if the settings key doesn't exist (unlikely), create it

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,1092 @@
...
+
+	return &holdoverPluginSettings{
+		LocalHoldoverTimeout:   uint(localHoldoverTimeout),
+		LocalMaxHoldoverOffset: uint(localMaxHoldoverOffset),
+		MaxInSpecOffset:        uint(maxInSpecOffset),
+	}, nil
+}
+
+// setPluginSettings updates the holdover plugin settings on the specified profile in the PtpConfig CR.
+func setPluginSettings(
```
klaskosk: see comment on getPluginSettings, thi could also be a `plugins.go` function

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,1092 @@
...
+				assertHoldoverOutOfSpecToFreerun(prometheusAPI, nodeName, ptpConfigWithTsc,
+					tscProfileIndex, upstreamIface, holdoverPluginSettingsWithOutOfSpec, timeout,
+					expectedClockClasses, clockClassChanges)
+			})
+	})
+})
+
+// assertHoldoverInSpecToLocked validates that after upstream clock loss the clock enters holdover-in-spec,
+// then recovers to locked when upstream is restored. No FREERUN events should be generated.
+func assertHoldoverInSpecToLocked(
```
klaskosk, line 226: I feel like we could figure out a way to make the number of parameters more manageable, but it'll depend a fair bit on the other comments

klaskosk (commented):
Thanks for getting the PR made! and for figuring out the issue with DNS, it'll be good to have that fix

there's 2 main themes to the comments:
- we should figure out how to classify the T-BC TR and T-TSC profiles in the main profile classification function
- a lot of the plugin/settings helpers could go into `profiles/plugins.go`. Maybe eco-goinfra too. we should fix the division in the future since it's unintended, but `plugins.go` works mainly on profiles and eco-goinfra works mainly on configs

tests/cnf/ran/ptp/internal/profiles/parser.go
dpopsuev: Fixed. Changed to `numInterfaces >= 1 && numServerInterfaces == numInterfaces`.

tests/cnf/ran/ptp/tests/ptp-holdover.go
dpopsuev: OC and T-TSC are fundamentally different at the protocol level: OC is IEEE 1588 standard PTP (can be slave or master, clock class 6), T-TSC is ITU-T G.8275.1 telecom profile (slave-only, clock class 255, mandatory hardware holdover). The [ADR gist](https://gist.github.com/dpopsuev/fec8bbf0f862087ad4b05fb4fb517fac#the-telecom-profile-itu-t-g82751) covers this.

Implemented `ProfileTypeTBCReceiver` and `ProfileTypeTTSC` in the parser. `isTimeSlaveClockProfile` deleted from the test file.

tests/cnf/ran/ptp/internal/profiles/parser.go
dpopsuev: Done. Added `ProfileTypeTBCReceiver` and `ProfileTypeTTSC` to the parser. `getControlledNamesForNode` computes controlled names in `GetNodeInfoMap`, passes context through to `determineProfileType`. Public API unchanged. `collectControlledProfileNames` and `isTimeSlaveClockProfile` deleted from the test file.

tests/cnf/ran/ptp/tests/ptp-holdover.go
dpopsuev: Fixed. Added `profiles.GetUpstreamPort` which uses the typed `IntelPlugin` struct and reads from `InputDelays[].UpstreamPort`. Removed the manual JSON parsing.

tests/cnf/ran/ptp/tests/ptp-holdover.go
dpopsuev: Done. Added `HoldoverPluginSettings`, `GetHoldoverPluginSettings`, and `SetHoldoverPluginSettings` to `profiles/plugins.go`. Read uses typed `IntelPlugin` struct via `getIntelPlugin`. Write uses raw JSON map to preserve unknown fields during round-trip.

tests/cnf/ran/ptp/tests/ptp-holdover.go
dpopsuev: Fixed. Replaced `getPluginName` with `GetPluginTypesFromProfile` via `isUnsupportedPlugin` helper that checks for `PluginTypeE825` and `PluginTypeE830` directly.

tests/cnf/ran/ptp/tests/ptp-holdover.go
dpopsuev: Done. `SetHoldoverPluginSettings` is in `profiles/plugins.go`. See comment #6.

tests/cnf/ran/ptp/internal/tsparams/consts.go
dpopsuev: Renamed to `LabelTBCTSCHoldover = "tbc-tsc-holdover"`.

tests/cnf/ran/ptp/tests/ptp-holdover.go
dpopsuev: Introduced `holdoverTestContext` struct grouping the 5 per-node params (prometheusAPI, nodeName, ptpConfig, profileIndex, upstreamIface) that are discovered once in BeforeEach. Top-level functions went from 9 params to 5. Sub-helpers keep their own smaller signatures.

dpopsuev:
@klaskosk I've been looking at the `parsePtpProfile` signature and I can see the seam where parsing and classification should split. I know you mentioned a parser overhaul is planned, so I'll hold off, but I wanted to flag that I've been thinking about it and I'm ready to pick it up whenever the time is right.

klaskosk:
> @klaskosk I've been looking at the `parsePtpProfile` signature and I can see the seam where parsing and classification should split. I know you mentioned a parser overhaul is planned, so I'll hold off, but I wanted to flag that I've been thinking about it and I'm ready to pick it up whenever the time is right.

@dpopsuev I want to finish the migration and get this version running in all our CI, but we should definitely meet to discuss the improvements (both from your doc and the parser) so we can plan them and add to our backlog

What I was mostly hinting at with the parser is that in 4.22, there's an HTTP endpoint that exposes the mapping of profiles to nodes that we should use. But seeing the code again as part of this PR, there's definitely other areas we could improve :sweat_smile:

dpopsuev:
@klaskosk Sure, I'm down for small iterations, waiting to hear if the code is good enough as is.

tests/cnf/ran/ptp/internal/profiles/plugins.go
```diff
@@ -154,3 +155,95 @@ func GetPluginTypesFromProfile(profile *ptpv1.PtpProfile) ([]ptp.PluginType, err
...
+		LocalHoldoverTimeout:   uint(localHoldoverTimeout),
+		LocalMaxHoldoverOffSet: uint(localMaxHoldoverOffset),
+		MaxInSpecOffset:        uint(maxInSpecOffset),
+	}, nil
+}
+
+// SetHoldoverPluginSettings patches the holdover plugin settings on the E810 plugin in the profile.
+// Uses raw JSON map to preserve unknown fields during round-trip.
+func SetHoldoverPluginSettings(profile *ptpv1.PtpProfile, settings HoldoverPluginSettings) error {
+	pluginJSON, pluginFound := profile.Plugins[string(ptp.PluginTypeE810)]
```
klaskosk: this could also use getIntelPlugin and then just json.Marshal the plugin, that way we're only parsing the plugin in one spot

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,766 @@
...
+
+	Context("t-bc upstream clock loss & unassisted holdover", func() {
+		var testCtx holdoverTestContext
+
+		timeout := holdoverTestTimeout
+
+		BeforeEach(func() {
+			By("getting node info map")
+
+			ctx := discoverHoldoverTestContext(prometheusAPI, profiles.ProfileTypeTBCReceiver)
```
klaskosk: minor: since `ctx` is so ingrained in Go as referring to `context.Context`, I'd prefer we stick with different names when referring to `holdoverTestContext`

perhaps even avoiding context as the name? `holdoverTestData` seems appropriate and avoids any confusion with the context package

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,766 @@
...
+			clockClassChanges    bool
+			expectedClockClasses holdoverExpectedClockClasses
+		)
+
+		timeout := holdoverTestTimeout
+
+		BeforeEach(func() {
+			is420, err := version.IsVersionStringInRange(
+				RANConfig.Spoke1OperatorVersions[ranparam.PTP], "4.20.0-0", "4.21.0-0")
+			Expect(err).ToNot(HaveOccurred())
```
klaskosk: this is a general thing, since I see a few places where we're using `Expect(err).ToNot(HaveOccurred())` without a description string. We should always provide a string here, that way it's clear from the error message itself what we were trying to do. I usually try to align with the `By`

for example, later on in `assertHoldoverInSpecToLocked`, you could do `Expect(err).ToNot(HaveOccurred(), "Failed to set upstream clock interface down to enter holdover-in-spec")`. this produces more descriptive error messages than something like "failed to set interface down" since it's clear what your goal within the test was

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,766 @@
...
+	}
+)
+
+// holdoverTestContext groups the per-node test context that is discovered once in BeforeEach and shared
+// by all test cases within a Context block.
+type holdoverTestContext struct {
+	prometheusAPI prometheusv1.API
+	nodeName      string
+	ptpConfig     *ptp.PtpConfigBuilder
+	profileIndex  int
```
klaskosk, line 50: I don't really like that we need to pass around the config builder and profile index instead of a `ProfileInfo`, but I've accepted that our helpers do a pretty bad job at elegantly allowing profile updates so is what it is. definitely an area we'll have to improve on later 🙂

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,766 @@
...
+
+	DeferCleanup(func() {
+		By("restoring original plugin settings")
+
+		restoreErr := profiles.SetHoldoverPluginSettings(profile, original)
+		Expect(restoreErr).ToNot(HaveOccurred(), "Failed to restore original plugin settings")
+
+		_, restoreErr = ctx.ptpConfig.Update()
+		Expect(restoreErr).ToNot(HaveOccurred(), "Failed to update PtpConfig with restored plugin settings")
+	})
```
klaskosk, line 707: I think we need to wait for the daemon to reload and lock again in the DeferCleanup. Otherwise, we can't guarantee that the state is clean again after each test

it's probably not a frequent issue, but it could cause a problem when the next test goes to update the profiles

klaskosk (commented):
a few minor things left, but that should be it. I went through the test flow in more detail and it checks out

tests/cnf/ran/ptp/internal/profiles/plugins.go
dpopsuev: Done.

tests/cnf/ran/ptp/tests/ptp-holdover.go
dpopsuev: Done.

tests/cnf/ran/ptp/tests/ptp-holdover.go
dpopsuev: Done.

tests/cnf/ran/ptp/tests/ptp-holdover.go
dpopsuev: Done.

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -0,0 +1,780 @@
...
+	return TBCClockClasses()
+}
+
+var _ = Describe("PTP Holdover", Label(tsparams.LabelTBCTSCHoldover), func() {
+	var prometheusAPI prometheusv1.API
+
+	BeforeEach(func() {
+		var err error
+
+		By("creating a Prometheus API client")
```
klaskosk, line 95: since we're not putting these tests into CI immediately and I know there's going to be a follow up PR we can just include it then, but we're missing a check in BeforeEach to ensure clocks are locked

```go
By("ensuring clocks are locked before testing")

err = metrics.EnsureClocksAreLocked(prometheusAPI)
Expect(err).ToNot(HaveOccurred(), "Failed to assert clock state is locked")
```

klaskosk (approved):
1 last comment, but I'd rather merge this so Hen can reuse the DPLL settings helper in the GM tests. Please be sure to include it in the PR for HardwareConfig/GNR-D though

PR #1383, opened 2026-05-08
CAPOA TLS profile tests

frajamomo:
- Add automated test suite (capoa-tls) validating that CAPOA bootstrap and controlplane controllers honor the cluster-wide TLS security profile from the APIServer CR
- Tests cover all four TLS profiles (Intermediate, Old, Modern, Custom), automatic reconciliation on profile changes, SecurityProfileWatcher restart behavior, webhook validation, and tlsAdherence policy changes
- Includes shared internal/tlsprofile/ helper package for TLS probing, port-forwarding, pod lifecycle, and APIServer patching — reusable by future component TLS suites (e.g., IBIO)
- Vendors k8s.io/client-go/tools/portforward for in-process port-forward support

 go vet passes
 golangci-lint passes
 All 9 test cases executed on OCP 4.22 nightly + MCE 2.17

## Summary by CodeRabbit

* **Tests**
  * Added a CAPOA TLS Profile end-to-end test suite with JUnit/XML reporting that validates TLS version/cipher acceptance/rejection, APIServer TLS profile transitions, webhook immutability, controller log assertions, pod restart/reconciliation, and cluster stability.
  * Added reusable test helpers, types, parameters, port‑forwarding utilities, TLS assertion helpers, pod lifecycle checks, and shared test init variables/configuration; suite-specific labels, namespaces, and reporter CRDs.

* **Chores**
  * Updated linter config to whitelist the test init package and add exclusions (including a depguard exclusion) for test packages.

### DIFF
--- a/tests/assisted/ztp/capoa-tls/internal/inittools/inittools.go
+++ b/tests/assisted/ztp/capoa-tls/internal/inittools/inittools.go
@@ -1,0 +1,28 @@
+package inittools
+
+import (
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/clients"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/internal/config"
+	"k8s.io/klog/v2"
+)
+
+var (
+	// HubAPIClient provides API access to hub cluster via KUBECONFIG env var.
+	HubAPIClient *clients.Settings
+	// GeneralConfig provides access to general configuration parameters.
+	GeneralConfig *config.GeneralConfig
+)
+
+func init() {
+	if GeneralConfig = config.NewConfig(); GeneralConfig == nil {
+		klog.Fatalf("failed to initialize GeneralConfig: config.NewConfig returned nil")
+	}
+
+	if HubAPIClient = clients.New(""); HubAPIClient == nil {
+		if GeneralConfig.DryRun {
+			return
+		}
+
+		klog.Fatalf("failed to initialize HubAPIClient: clients.New returned nil")
+	}
+}
--- a/tests/assisted/ztp/capoa-tls/internal/tsparams/capoa-tls-vars.go
+++ b/tests/assisted/ztp/capoa-tls/internal/tsparams/capoa-tls-vars.go
@@ -1,0 +1,23 @@
+package tsparams
+
+import (
+	"github.com/openshift-kni/k8sreporter"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/assisted/ztp/internal/ztpparams"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var (
+	// Labels represents the range of labels that can be used for test cases selection.
+	Labels = append(ztpparams.Labels, LabelSuite)
+
+	// ReporterNamespacesToDump tells to the reporter from where to collect logs.
+	ReporterNamespacesToDump = map[string]string{
+		"multicluster-engine": "mce",
+		TestNamespace:         "capoa-tls webhook test namespace",
+	}
+
+	// ReporterCRDsToDump tells to the reporter what CRs to dump.
+	ReporterCRDsToDump = []k8sreporter.CRData{
+		{Cr: &corev1.PodList{}},
+	}
+)
--- a/tests/assisted/ztp/capoa-tls/tests/capoa_tls_profile.go
+++ b/tests/assisted/ztp/capoa-tls/tests/capoa_tls_profile.go
@@ -1,0 +1,489 @@
+package capoa_tls_test
+
+import (
+	"context"
+	"crypto/tls"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/apiservers"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/clients"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/namespace"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/pod"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/reportxml"
+	. "github.com/rh-ecosystem-edge/eco-gotests/tests/assisted/ztp/capoa-tls/internal/inittools"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/assisted/ztp/capoa-tls/internal/tsparams"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/assisted/ztp/internal/tlsprofile"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var customCiphers = []string{
+	"ECDHE-RSA-AES128-GCM-SHA256",
+	"ECDHE-RSA-AES256-GCM-SHA384",
+	"ECDHE-ECDSA-AES128-GCM-SHA256",
+	"ECDHE-ECDSA-AES256-GCM-SHA384",
+}
+
+func customTLSProfile(ciphers []string) configv1.TLSSecurityProfile {
+	return configv1.TLSSecurityProfile{
+		Type: configv1.TLSProfileCustomType,
+		Custom: &configv1.CustomTLSProfile{
+			TLSProfileSpec: configv1.TLSProfileSpec{
+				Ciphers:       ciphers,
+				MinTLSVersion: configv1.VersionTLS12,
+			},
+		},
+	}
+}
+
+var capoa = &tlsprofile.Component{
+	Name:        "CAPOA",
+	Namespace:   "multicluster-engine",
+	RestartMode: tlsprofile.RestartModeContainerRestart,
+	Endpoints: []tlsprofile.Endpoint{
+		{
+			ServiceName:    "capoa-bootstrap-webhook-service",
+			LocalPort:      19443,
+			RemotePort:     9443,
+			DeploymentName: "capoa-bootstrap-controller-manager",
+		},
+		{
+			ServiceName:    "capoa-controlplane-webhook-service",
+			LocalPort:      19444,
+			RemotePort:     9443,
+			DeploymentName: "capoa-controlplane-controller-manager",
+		},
+	},
+	Deployments: []tlsprofile.Deployment{
+		{Name: "capoa-bootstrap-controller-manager", ContainerName: "manager"},
+		{Name: "capoa-controlplane-controller-manager", ContainerName: "manager"},
+	},
+	ListPods: func(client *clients.Settings, ns string) ([]*pod.Builder, error) {
+		return pod.ListByNamePattern(client, "capoa", ns)
+	},
+	ExpectedHealthyPods: 2,
+	PodReadyTimeout:     5 * time.Minute,
+	AutoRestartTimeout:  10 * time.Minute,
+	HonoringLogPattern:  "honoring cluster-wide TLS profile",
+	AllowedCipher:       tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+	AllowedCipherAlt:    tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+	DisallowedCipher:    tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,
+	OldProfileCipher:    tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
+}
+
+const defaultsLogPattern = "using defaults"
+
+func ensureTLSAdherence() {
+	builder, err := apiservers.PullAPIServer(HubAPIClient)
+	Expect(err).ToNot(HaveOccurred(), "failed to pull apiserver/cluster")
+
+	if builder.Definition.Spec.TLSAdherence == configv1.TLSAdherencePolicyStrictAllComponents {
+		return
+	}
+
+	By("Setting tlsAdherence to StrictAllComponents")
+	tlsprofile.PatchTLSAdherence(HubAPIClient, configv1.TLSAdherencePolicyStrictAllComponents)
+}
+
+// Tests are ordered to minimize TLS profile changes and cluster churn.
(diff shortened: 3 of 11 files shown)


tests/assisted/ztp/internal/tlsprofile/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			return fmt.Sprintf("failed to list cluster operators: %v", err)
+		}
+
+		for i := range coList.Items {
+			co := &coList.Items[i]
+			available := false
+			progressing := true
+			degraded := true
+
```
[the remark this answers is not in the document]
frajamomo: The current defaults are intentionally pessimistic and correct. If a ClusterOperator is missing its Progressing or Degraded condition entirely, that means the operator hasn't reported its status yet, which is an unstable state that should not pass the stability check. Defaulting to progressing=false, degraded=false would treat an operator with no conditions as "stable", which could allow tests to proceed while the operator is still initializing.

In practice, all OCP cluster operators always report all three conditions — this is a Kubernetes convention enforced by the CVO. The pessimistic defaults are a safety net for edge cases (operator just deployed, conditions not yet populated), which is the correct behavior for a stability gate.

rccrdpccl:
As CAPOA runs slightly differently on OCP vs vanilla k8s  it's important to make sure we test edge cases around possible scenarios where the controller could get into "k8s mode" (with no API server tls API)

rccrdpccl:
What happens when the API (OCP's tls profile) returns a transient error? Do we have any way to inject these types of errors?

tests/assisted/ztp/internal/tlsprofile/helpers.go
```diff
@@ -0,0 +1,102 @@
...
+			"tlsSecurityProfile": profile,
+		},
+	}
+
+	patchBytes, err := json.Marshal(patchMap)
+	Expect(err).ToNot(HaveOccurred(), "failed to marshal TLS profile patch")
+
+	apiserver := &configv1.APIServer{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}}
+	err = client.Patch(context.TODO(), apiserver, runtimeclient.RawPatch(types.MergePatchType, patchBytes))
+	Expect(err).ToNot(HaveOccurred(), "failed to patch apiserver TLS profile")
```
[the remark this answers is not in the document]
frajamomo: The Patch calls here use RawPatch(types.MergePatchType, ...), which is a server-side merge patch — it does not read-modify-write, so it cannot produce a conflict error (IsConflict). The patch is applied atomically by the API server.

These are test setup helpers that run before any rollout begins (they trigger the rollout). There's no concurrent control-plane activity modifying the same APIServer object at the time the patch is applied. The only realistic transient failure would be a momentary API server unavailability, but:

The test framework already has Eventually-based retries at the caller level for validating state after the patch.
Adding wait.ExponentialBackoff would add complexity (new imports, retry logic, error classification) to a helper that is called once per test setup, and the failure mode it addresses (transient 429/timeout on a single Patch during test setup) is extremely unlikely.
If the API server is genuinely down at test setup time, retrying a few times with backoff won't fix the underlying problem — the test environment is broken.
No changes made. The current code is appropriate for its test-helper context.

tests/assisted/ztp/capoa-tls/internal/tsparams/capoa-tls-vars.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/openshift-kni/k8sreporter"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/assisted/ztp/internal/ztpparams"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var (
+	// Labels represents the range of labels that can be used for test cases selection.
+	Labels = append(ztpparams.Labels, LabelSuite)
```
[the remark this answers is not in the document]
frajamomo: The source slice ztpparams.Labels has len == cap == 2 (literal initialization), so append(..., LabelSuite) always allocates a new backing array. There is no mutation risk, and all three test suites in the repo use the same pattern. No change needed.

tests/assisted/ztp/capoa-tls/internal/inittools/inittools.go
```diff
@@ -0,0 +1,18 @@
...
+	// HubAPIClient provides API access to hub cluster via KUBECONFIG env var.
+	HubAPIClient *clients.Settings
+	// GeneralConfig provides access to general configuration parameters.
+	GeneralConfig *config.GeneralConfig
+)
+
+func init() {
+	HubAPIClient = clients.New("")
+	GeneralConfig = config.NewConfig()
+}
```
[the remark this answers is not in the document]
frajamomo: Fix applied and validated

frajamomo:
Fix applied and validated. Removed 6 redundant WaitPodsReady calls that immediately followed WaitPodsRestarted — at lines 328, 336, 428, 439, 457, and 490. WaitPodsRestarted already calls WaitPodsReady internally ([pods.go:75](vscode-webview://0mko502ls92t1htrnpk8hvj3ipepm2btgj9pa3267ig71j61a78p/tests/assisted/ztp/internal/tlsprofile/pods.go#L75)), so the duplicates just added unnecessary wait time. The standalone WaitPodsReady call on line 480 (test 89004, checking baseline from previous test without a restart) was kept as it serves a different purpose.

tests/assisted/ztp/internal/tlsprofile/tls.go
```diff
@@ -0,0 +1,105 @@
...
+		if dialErr == nil {
+			return "connected"
+		}
+
+		errMsg := dialErr.Error()
+		if strings.Contains(errMsg, "connection refused") ||
+			strings.Contains(errMsg, "EOF") ||
+			strings.Contains(errMsg, "connection reset") {
+			return "not-ready"
+		}
```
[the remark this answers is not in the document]
frajamomo: Fixed

frajamomo:
> As CAPOA runs slightly differently on OCP vs vanilla k8s it's important to make sure we test edge cases around possible scenarios where the controller could get into "k8s mode" (with no API server tls API)

Testing addressed in [Add unit tests for OCP vs vanilla K8s mode edge cases](https://github.com/openshift-assisted/cluster-api-provider-openshift-assisted/pull/718#top)

frajamomo:
@coderabbitai review

tests/assisted/ztp/capoa-tls/tests/capoa_tls_profile.go
```diff
@@ -0,0 +1,509 @@
...
+
+const (
+	adherenceStrictAllComponents = "StrictAllComponents"
+	adherenceLegacyAdheringOnly  = "LegacyAdheringComponentsOnly"
+
+	defaultsLogPattern = "using defaults"
+)
+
+func ensureTLSAdherence() {
+	apiserverU := &unstructured.Unstructured{}
```
klaskosk: rather than using unstructured, I would recommend creating a builder in eco-goinfra. There's already a `pkg/apiservers` there, although that doesn't seem to have the `config.openshift.io` one. So I'd recommend creating one there to follow the typical split of eco-gotests and eco-goinfra

tests/assisted/ztp/internal/tlsprofile/helpers.go
```diff
@@ -0,0 +1,102 @@
...
+
+	err := client.Patch(context.TODO(), apiserver,
+		runtimeclient.RawPatch(types.MergePatchType, patchBytes))
+	Expect(err).ToNot(HaveOccurred(), "failed to remove apiserver TLS profile")
+}
+
+// WaitForClusterStability waits until all cluster operators are Available and not
+// Progressing or Degraded. This prevents test cases from running while the control
+// plane is mid-rollout from a previous TLS profile change.
+func WaitForClusterStability(client *clients.Settings, timeout time.Duration) {
```
klaskosk, line 47: in eco-goinfra `pkg/clusteroperator`, there's a `WaitForAllClusteroperatorsAvailable` that could be reused instead of creating a separate function here

klaskosk (commented):
ultimately it's up to whoever owns the assisted/ztp directory (@trewest I think? based on git history) so I won't comment too much, but some of the infra helpers would be more idiomatic in eco-goinfra

also not sure if it's possible given the CI workflows and all, but [openshift/tls-scanner](https://github.com/openshift/tls-scanner) provides a way to do a lot of TLS adherence checks

tests/assisted/ztp/internal/tlsprofile/helpers.go
frajamomo: The existing eco-goinfra functions don't cover the same semantics:

- `WaitForAllClusteroperatorsAvailable` only checks `Available=True`

- `WaitForAllClusteroperatorsStopProgressing` only checks `Progressing` — and returns false (not progressing) when the condition is missing, while we intentionally assume `progressing=true` for missing conditions

Neither checks `Degraded=False`
We need all three conditions checked atomically per operator with pessimistic defaults for missing conditions. Happy to contribute a combined `WaitForAllClusteroperatorsStable()` to eco-goinfra as a follow-up if that's useful.

klaskosk: fair enough :smile:

I assumed Available=True is mutually exclusive with Degraded=False, but if that's the case no worries

tests/assisted/ztp/capoa-tls/tests/capoa_tls_profile.go
frajamomo: I looked into this and created a `config.openshift.io/v1` APIServer builder in eco-goinfra ([frajamomo/eco-goinfra@apiserver-config-builder](https://github.com/frajamomo/eco-goinfra/tree/apiserver-config-builder)) following the DNS/Proxy/Infrastructure singleton pattern — PullAPIServer, Get, Exists, Update, WithTLSSecurityProfile.

However, when I wired it into these tests I hit a blocker: the builder's `Update()` does a full object replacement, which drops `spec.tlsAdherence` — a field behind the TLSAdherence feature gate that isn't in the vendored configv1.APIServer struct. The API server rejects this with "tlsAdherence may not be removed once set".

The current raw MergePatch approach only touches the fields it explicitly sets, so it coexists safely with feature-gated fields. To make the builder work here, it would need a patch-based update method (e.g. MergePatch or server-side apply) rather than a full Update. That's a pattern I don't see in eco-goinfra today, so it would need some design discussion.

For now I'll keep the raw patches in the helper functions. The builder PR is ready if/when eco-goinfra adds patch support, or if the tlsAdherence field graduates to the stable API — at that point a typed Update() would work fine.

klaskosk: we've been having a few other people (including myself :slightly_smiling_face:) run into issues with the dependencies not being updated, so I plan on merging the updates that'll give you the field today

frajamomo:
[openshift/tls-scanner](https://github.com/openshift/tls-scanner) is a different tool for a different purpose. 

tls-scanner does a broad cluster-wide port scan (every pod, every TCP port) and produces a compliance report. The CAPOA tests do something more targeted: they verify that specific CAPOA webhook endpoints respond correctly to each TLS profile change in real time (connect with allowed cipher → success, connect with disallowed cipher → rejection), verify controller restart behavior, and check log messages. 

tls-scanner wouldn't cover the reconciliation/restart/webhook-validation aspects of these tests.

frajamomo:
Now depends on https://github.com/rh-ecosystem-edge/eco-goinfra/pull/1518 @klaskosk . I'll keep this one in draft until the eco-goinfra MR is merged.

frajamomo:
Depends on https://github.com/rh-ecosystem-edge/eco-gotests/pull/1454 to access new eco-goinfra code referenced

frajamomo:
Depends also on the https://github.com/rh-ecosystem-edge/eco-goinfra/pull/1519 for eco-goinfra to provide portForwarding support code

frajamomo:
Depend on https://github.com/rh-ecosystem-edge/eco-gotests/pull/1460 to update eco-goinfra reference

frajamomo:
Rebased to pick up the supporting functions in eco-goinfra. Ready for review. @klaskosk

tests/assisted/ztp/internal/tlsprofile/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/apiservers"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/clients"
+)
+
+// PatchAPIServerTLSProfile applies the given TLS security profile to the cluster APIServer.
+func PatchAPIServerTLSProfile(client *clients.Settings, profile configv1.TLSSecurityProfile) {
```
klaskosk: nit: now that it's a regular update, the Patch naming is no longer accurate

tests/assisted/ztp/capoa-tls/tests/capoa_tls_profile.go
```diff
@@ -0,0 +1,489 @@
...
+				_, err := nsBuilder.Create()
+				Expect(err).ToNot(HaveOccurred(), "failed to create test namespace")
+
+				DeferCleanup(func() {
+					_ = nsBuilder.DeleteAndWait(2 * time.Minute)
+				})
+
+				By("Creating valid OpenshiftAssistedConfig")
+
+				resource := &unstructured.Unstructured{
```
klaskosk, line 261: it'd also be more idiomatic to use an eco-goinfra builder here, but not that important unless you start needing to use OpenshiftAssistedConfig in other places

klaskosk (commented):
thanks for your eco-goinfra contributions! I'm fine with this PR but whoever owns `assisted/ztp` needs to provide the final approval here

there's only 1 approval needed to merge so once they add the approval you should be good, though let me know if you still have trouble merging and we can work that out

tests/assisted/ztp/internal/tlsprofile/helpers.go
frajamomo: Renamed PatchAPIServerTLSProfile → SetAPIServerTLSProfile — the function uses the builder Update() now, not a raw patch. Updated all 6 call sites.

tests/assisted/ztp/capoa-tls/tests/capoa_tls_profile.go
frajamomo: It's used in one place only, and there's no eco-goinfra builder for this CRD. Not worth the abstraction for a single test helper.

trewest (approved):

PR #1392, opened 2026-05-14
[PTP] Migrate T-GM GNSS loss tests from cnf-gotests

hhassid:
## Summary
- Migrate Polarion IDs 78463, 78464, and 78465 from `cnf-gotests` (`test/ran/ptp/tests/ptp_recovery.go`) to `eco-gotests`
- New test file `tests/cnf/ran/ptp/tests/ptp-gnss-loss.go` with three test cases validating T-GM clock state transitions during GNSS loss:
  - **78463**: holdover → locked (GNSS recovery before timeout)
  - **78464**: holdover → freerun (holdover timeout expiration)
  - **78465**: holdover → freerun (DPLL offset exceeds max in-spec)
- Added `LabelGNSSLoss` constant to `tests/cnf/ran/ptp/internal/tsparams/consts.go`
- Tests support both SNO and MNO clusters — on MNO, the test runs once on the first node with a GM profile

**Story:** https://redhat.atlassian.net/browse/CNF-20471

**Note:** This PR was written with the assistance of Cursor AI.

## Test plan
- [x] Run against a cluster with T-GM (Grandmaster) PTP configuration and GNSS-capable NICs - WPC
- [ ] Run against a cluster with T-GM (Grandmaster) PTP configuration and GNSS-capable NICs - GNR-D

## Summary by CodeRabbit

* **Tests**
  * Added a GNSS-loss test suite with three scenarios validating PTP transitions, holdover timing, out-of-spec offset handling, Prometheus metric checks, daemon log validations, and a test ensuring no FREERUN event is emitted after recovery.
* **Chores**
  * Added a label for GNSS-loss test tagging and an event-time filter to improve event-based test waits.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/eco-gotests/pull/1392?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
+++ b/tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
@@ -1,0 +1,716 @@
+package tests
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	eventptp "github.com/redhat-cne/sdk-go/pkg/event/ptp"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/ptp"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/reportxml"
+	"k8s.io/klog/v2"
+
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/internal/querier"
+	. "github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/internal/raninittools"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/internal/ranparam"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/internal/version"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/consumer"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/daemonlogs"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/events"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/gnss"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/iface"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/metrics"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/processes"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/profiles"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/tsparams"
+)
+
+var _ = Describe("PTP T-GM GNSS Loss", Label(tsparams.LabelGNSSLoss), func() {
+	var (
+		prometheusAPI   prometheusv1.API
+		savedPtpConfigs []*ptp.PtpConfigBuilder
+		configSupported bool
+	)
+
+	eventTimeout := 5 * time.Minute
+
+	BeforeEach(func() {
+		By("skipping if PTP version is below 4.18")
+
+		inRange, err := version.IsVersionStringInRange(
+			RANConfig.Spoke1OperatorVersions[ranparam.PTP], "4.18.0-0", "")
+		Expect(err).ToNot(HaveOccurred(), "Failed to check PTP version range")
+
+		if !inRange {
+			Skip("Test is valid from PTP version 4.18 and higher")
+		}
+
+		By("creating a Prometheus API client")
+
+		prometheusAPI, err = querier.CreatePrometheusAPIForCluster(RANConfig.Spoke1APIClient)
+		Expect(err).ToNot(HaveOccurred(), "Failed to create Prometheus API client")
+
+		By("ensuring clocks are locked before testing")
+
+		err = metrics.EnsureClocksAreLocked(prometheusAPI)
+		Expect(err).ToNot(HaveOccurred(), "Failed to assert clock state is locked")
+
+		By("saving PtpConfigs before testing")
+
+		savedPtpConfigs, err = profiles.SavePtpConfigs(RANConfig.Spoke1APIClient)
+		Expect(err).ToNot(HaveOccurred(), "Failed to save PtpConfigs")
+
+		configSupported, err = version.IsVersionStringInRange(
+			RANConfig.Spoke1OperatorVersions[ranparam.PTP], "4.20.0-0", "")
+		Expect(err).ToNot(HaveOccurred(), "Failed to check PTP config label version range")
+	})
+
+	AfterEach(func() {
+		By("restoring PtpConfigs after testing")
+
+		startTime := time.Now()
+		changedProfiles, err := profiles.RestorePtpConfigs(RANConfig.Spoke1APIClient, savedPtpConfigs)
+		Expect(err).ToNot(HaveOccurred(), "Failed to restore PtpConfigs")
+
+		if len(changedProfiles) > 0 {
+			By("waiting for profile load on nodes")
+
+			err := daemonlogs.WaitForProfileLoadOnPTPNodes(RANConfig.Spoke1APIClient,
+				daemonlogs.WithStartTime(startTime),
+				daemonlogs.WithTimeout(5*time.Minute))
+			if err != nil {
+				klog.V(tsparams.LogLevel).Infof("Failed to wait for profile load on PTP nodes: %v", err)
+			}
+		}
+
+		By("ensuring clocks are locked after testing")
+
+		err = metrics.EnsureClocksAreLocked(prometheusAPI)
+		Expect(err).ToNot(HaveOccurred(), "Failed to assert clock state is locked")
+	})
+
+	// 78463 - verifies t-gm transition from holdover to locked due to gnss recovery
+	It("verifies t-gm transition from holdover to locked due to gnss recovery",
+		reportxml.ID("78463"), func() {
+			testActuallyRan := false
+
+			By("getting node info map")
+
+			nodeInfoMap, err := profiles.GetNodeInfoMap(RANConfig.Spoke1APIClient)
+			Expect(err).ToNot(HaveOccurred(), "Failed to get node info map")
+
+			for nodeName, nodeInfo := range nodeInfoMap {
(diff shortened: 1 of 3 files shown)


tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -0,0 +1,689 @@
...
+		err := gnss.SimulateSyncRecovery(RANConfig.Spoke1APIClient, nodeName, protocolVersion)
+		if err != nil {
+			klog.V(tsparams.LogLevel).Infof("Failed to restore GNSS sync on node %s: %v", nodeName, err)
+		}
+	}
+}
+
+// setPluginDPLLSettings modifies the DPLL settings in the GM profile's Intel plugin configuration. The settings map
+// keys should be the DPLL setting names (e.g. "LocalHoldoverTimeout", "MaxInSpecOffset", "LocalMaxHoldoverOffset").
+func setPluginDPLLSettings(gmProfile *profiles.ProfileInfo, settings map[string]uint64) error {
```
klaskosk: I suspect #1381 will merge before this, in which case we'll be able to reuse the helpers Daniel added for DPLL settings. although his first PR is focused on e810, so it may need updated a bit

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -0,0 +1,689 @@
...
+
+			if !testActuallyRan {
+				Skip("Test requires Grandmaster configuration")
+			}
+		})
+})
+
+// cleanupGNSSLoss returns a cleanup function that restores GNSS sync on a node if the test failed. This is intended
+// to be used with DeferCleanup to ensure GNSS is restored even when a test assertion fails mid-way.
+func cleanupGNSSLoss(nodeName, protocolVersion string) func() {
```
klaskosk: this is pretty much the same as `cleanupGNSSSync` from `ptp-ntp-fallback.go`, but it raises a good point that it'd be violating the rule of keeping unexported functions within a file to reuse that function

the main difference is that `cleanupGNSSSync` also checks for `ts2phc` lock, which I'd recommend doing here too since it's important to ensure the process is fully recovered before moving on

I don't think this would fit nicely in a `internal/` package, so maybe we drop the rule of keeping unexported functions within a file? I'm curious what you think (and maybe @dpopsuev has some thoughts here too)

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -0,0 +1,689 @@
...
+				err = setPluginDPLLSettings(gmProfile, map[string]uint64{
+					dpllSettingLocalHoldoverTimeout:   14400,
+					dpllSettingLocalMaxHoldoverOffset: 1500,
+					dpllSettingMaxInSpecOffset:        1000,
+				})
+				Expect(err).ToNot(HaveOccurred(), "Failed to set plugin DPLL settings on node %s", nodeName)
+
+				By("waiting for profile load after config change")
+
+				configChangeTime := time.Now()
```
klaskosk: this line should go right before the setPluginDPLLSettings so that we're not missing the log in case it came in between the functions

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -0,0 +1,689 @@
...
+				err = setPluginDPLLSettings(gmProfile, map[string]uint64{
+					dpllSettingLocalHoldoverTimeout:   holdoverTimeoutSeconds,
+					dpllSettingLocalMaxHoldoverOffset: 300,
+					dpllSettingMaxInSpecOffset:        600,
+				})
+				Expect(err).ToNot(HaveOccurred(), "Failed to set plugin DPLL settings on node %s", nodeName)
+
+				By("waiting for profile load after config change")
+
+				configChangeTime := time.Now()
```
klaskosk: same as above

> this line should go right before the setPluginDPLLSettings so that we're not missing the log in case it came in between the functions

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -0,0 +1,689 @@
...
+				err = setPluginDPLLSettings(gmProfile, map[string]uint64{
+					dpllSettingLocalHoldoverTimeout:   holdoverTimeoutSeconds,
+					dpllSettingLocalMaxHoldoverOffset: localMaxHoldoverOffset,
+					dpllSettingMaxInSpecOffset:        maxInSpecOffset,
+				})
+				Expect(err).ToNot(HaveOccurred(), "Failed to set plugin DPLL settings on node %s", nodeName)
+
+				By("waiting for profile load after config change")
+
+				configChangeTime := time.Now()
```
klaskosk: and here as well

> this line should go right before the setPluginDPLLSettings so that we're not missing the log in case it came in between the functions

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -0,0 +1,689 @@
...
+					dpllSettingLocalMaxHoldoverOffset: 1500,
+					dpllSettingMaxInSpecOffset:        1000,
+				})
+				Expect(err).ToNot(HaveOccurred(), "Failed to set plugin DPLL settings on node %s", nodeName)
+
+				By("waiting for profile load after config change")
+
+				configChangeTime := time.Now()
+
+				err = daemonlogs.WaitForProfileLoadOnPTPNodes(RANConfig.Spoke1APIClient,
```
klaskosk: I forget if this gets addressed by using the helper from Daniel's PR, but we should only wait for profile load if we know the DPLL settings changed

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -0,0 +1,689 @@
...
+				err = events.WaitForEvent(eventPod, gpsLossTime, eventTimeout, events.All(
+					events.IsType(eventptp.PtpClockClassChange),
+					events.HasValue(events.WithMetric(int64(metrics.ClockClass6))),
+				))
+				Expect(err).ToNot(HaveOccurred(), "Failed to receive clock class 6 event on node %s", nodeName)
+
+				By("verifying clock class 6 in metrics")
+
+				clockClassQuery := metrics.ClockClassQuery{
+					Process: metrics.Equals(metrics.ProcessPTP4L),
```
klaskosk, line 243: I see in cnf-gotests we're specifying the config for this query for 4.20+, why are we not here?

It seems fine because there's no harm in verifying all ptp4l processes are clock class 6, but I forget why exactly we were specific in cnf-gotests

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -0,0 +1,689 @@
...
+		})
+
+	// 78465 - verifies t-gm transition from holdover to freerun due to offset
+	It("verifies t-gm transition from holdover to freerun due to offset",
+		reportxml.ID("78465"), func() {
+			const (
+				holdoverTimeoutSeconds   = 600
+				localMaxHoldoverOffset   = 60000
+				maxInSpecOffset          = 300
+				timeToReachMaxInSpecSecs = maxInSpecOffset / (localMaxHoldoverOffset / holdoverTimeoutSeconds)
```
klaskosk: nit: we could multiply by `time.Second` when creating the constant so it's clear from the beginning the units

klaskosk (commented):
overall the tests look great, main comment is the ordering of the profile load times and ensuring we're not waiting for a profile load if no changes occur

Daniel's PR will add a helper for the DPLL settings that I'd prefer we use, but it's not nearly merged yet but should very soon

There's two questions too, nothing huge though

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
hhassid: Using GetHoldoverPluginSettings now

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
hhassid: For versions 4.20+ the slaves clock-class is 255 and not 6.
the config filed is only added for 4.20+.
Config added to the clockClassQuery

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
hhassid: Fixed.

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
hhassid: I get your point.
I think or move it to internal/gnss
or duplicate it (yuck)
Anyway I think we should keep the rule

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -0,0 +1,734 @@
...
+
+				By("waiting for clock class 7 event")
+
+				err = events.WaitForEvent(eventPod, gpsLossTime, eventTimeout, events.All(
+					events.IsType(eventptp.PtpClockClassChange),
+					events.HasValue(events.WithMetric(int64(metrics.ClockClass7))),
+				))
+				Expect(err).ToNot(HaveOccurred(), "Failed to receive clock class 7 event on node %s", nodeName)
+
+				By("waiting for os-clock-sync-state FREERUN event for CLOCK_REALTIME")
```
klaskosk: AI pointed out that this check doesn't occur in the cnf-gotests version. It's good to include the extra check here though, as long as we know it's what the behavior of the actual cluster is. So assuming this passes testing then it's a good thing 🙂

klaskosk (dismissed):

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
klaskosk: I guess we ended up going with just reusing the function, that works for now. addressing setup/restore boilerplate is high on the priorities for once we finish the initial migration so we'll have the official stance then

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -0,0 +1,734 @@
...
+
+				profileToUpdate := &ptpConfig.Definition.Spec.Profile[gmProfileInfo.Reference.ProfileIndex]
+
+				desiredSettings := profiles.HoldoverPluginSettings{
+					LocalHoldoverTimeout:   14400,
+					LocalMaxHoldoverOffSet: 1500,
+					MaxInSpecOffset:        1000,
+				}
+
+				currentSettings, err := profiles.GetHoldoverPluginSettings(profileToUpdate)
```
klaskosk, line 133: I'm realizing now that this doesn't work for GNR-D. But I think Daniel's already committed to addressing that case with the HardwareConfig CR in his PR so we can wait until then

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
hhassid: Right, it's a validation we should add for some of the tests https://redhat.atlassian.net/browse/CNF-23735 when the gnss is not available

klaskosk (approved):

PR #248, opened 2024-10-09
cnf network: fix rdma flake

ajaggapa:
MCP update is happening ~12 seconds while deleting an MC

PR #1409, opened 2026-05-26
Fix test runner pkg path discovery

zhiqiangf:
## Summary

When `ECO_GOTESTS_PACKAGE_PATH` is set in the Jenkins pipeline, the package
path (e.g. `./tests/ocp/sriov`) is passed as a positional argument to the
container. At the same time, `ECO_TEST_FEATURES=sriov` causes `test-runner.sh`
to discover all directories named `sriov` via `find`, which returns both
`./tests/ocp/sriov` **and** `./tests/cnf/core/network/sriov`. These get
appended to the ginkgo command via `$feature_dirs`, so the final invocation
ends up as:

```
ginkgo ... ./tests/ocp/sriov ./tests/cnf/core/network/sriov ./tests/ocp/sriov
```

`./tests/ocp/sriov` runs twice, doubling the test runtime (observed in build #248: 4h10m instead of ~2h).

## Fix

Allow `ECO_TEST_FEATURES` to be unset. When it is unset, `feature_dirs` is
left empty and the script relies solely on the explicit package path(s) passed
as positional arguments via `$@`. The paired Jenkins pipeline change skips
adding `ECO_TEST_FEATURES` to the container env when `ECO_GOTESTS_PACKAGE_PATH`
is already set.

## Test plan

- [ ] Run pipeline with `ECO_GOTESTS_PACKAGE_PATH=./tests/ocp/sriov` — verify ginkgo is invoked with that path only (no duplicate, no `cnf` suite)
- [ ] Run pipeline with `ECO_GOTESTS_PACKAGE_PATH` cleared — verify `ECO_TEST_FEATURES`-based discovery still works as before

## Summary by CodeRabbit

* **Bug Fixes**
  * Test runner no longer requires an environment variable to be set; tests can now run with explicit package path arguments even when the variable is unset.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/eco-gotests/pull/1409?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


zhiqiangf:
Closing in favor of keeping the fix on the Jenkins pipeline side only. The test-runner.sh change is too broad — many teams depend on ECO_TEST_FEATURES being required, and changing that behavior would need wider coordination. The paired Jenkins change (skipping ECO_TEST_FEATURES when ECO_GOTESTS_PACKAGE_PATH is set) is sufficient to fix the double-run.

PR #1411, opened 2026-05-27
[release-4.20] hw-accel: fix the in-tree replacement module name

ItzikEzra-rh:
## Summary

Cherry-pick of #967 to `release-4.20`.

Fixes the in-tree replacement test (test_id:62745) which has been consistently failing due to a module name mismatch introduced when build arguments were updated.

## Changes

One-line fix: use the correct module name so the built `.ko` file matches what modprobe expects to load.

## Testing

- Fix verified on `main` — test passes
- `release-4.20` currently fails 100% of runs for this test


cdvultur (approved):
/lgtm

PR #1413, opened 2026-05-27
CNF-24028: E825 HardwareConfig CRD support for GNR-D T-BC holdover tests

dpopsuev:
Closes https://redhat.atlassian.net/browse/CNF-24028

Depends on https://github.com/rh-ecosystem-edge/eco-goinfra/pull/1494

## Summary

- Add `profiles/hardwareconfig.go`: resolve `HardwareConfig` CR by `relatedPtpProfileName`, read/write `HoldoverParameters`
- Add `GetUpstreamPortForProfile` dispatcher: reads `PtpSettings["upstreamPort"]` (mandatory for GNRD T-BC, covers both plugin and HardwareConfig paths)
- Add `WaitForHardwareConfigLoad` in `daemonlogs`: waits for `"Successfully applied profile"` — the HardwareConfig restart path equivalent of `WaitForProfileLoad`
- Extend `ProfileInfo` with `HardwareConfig` field populated during `GetNodeInfoMap` via a single `ListHardwareConfigs` call
- Update `ptp-holdover.go`: remove E825 skip guard, dispatch get/apply holdover settings per path

## Test plan

- [ ] `go build ./tests/cnf/ran/ptp/...`
- [ ] `golangci-lint run ./tests/cnf/ran/ptp/...`
- [ ] 83298 (holdover-in-spec → freerun) PASSED on helix98 E825 GNR-D T-BC with HardwareConfig CR applied

## Summary by CodeRabbit

* **New Features**
  * Added HardwareConfig support for associating hardware configuration with PTP profiles
  * Extended holdover settings management with dual paths: hardware configuration and plugin-based
  * Added support for multiple Intel DPLL plugin types in precedence order (E810, E825, E830)
  * Improved upstream port resolution with profile-first and E810 plugin fallback mechanisms
  * Enhanced logging and acknowledgement waiting for hardware configuration application


tests/cnf/ran/ptp/internal/profiles/plugins.go
```diff
@@ -222,6 +222,21 @@ func SetHoldoverPluginSettings(profile *ptpv1.PtpProfile, settings HoldoverPlugi
...
+			return iface.Name(port), nil
+		}
+	}
+
+	return GetUpstreamPort(profile)
+}
+
 // GetUpstreamPort returns the upstream port from the E810 plugin's interconnections. Returns the first
 // interconnection entry that has an upstreamPort set.
 func GetUpstreamPort(profile *ptpv1.PtpProfile) (iface.Name, error) {
```
klaskosk: could you please rename this something like `GetUpstreamPortFromE810Plugin` so it's clearer the difference between this function and `GetUpstreamPortForProfile`?

tests/cnf/ran/ptp/internal/profiles/hardwareconfig.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/clients"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/ptp"
+	ptpv2alpha1 "github.com/rh-ecosystem-edge/eco-goinfra/pkg/schemes/ptp/v2alpha1"
+)
+
+// GetHardwareConfigForProfile returns the HardwareConfig CR whose RelatedPtpProfileName
+// matches profileName. The relationship is one-directional: HardwareConfig points to
+// PtpProfile, not the reverse. Returns an error if no matching CR is found.
+func GetHardwareConfigForProfile(
```
klaskosk: appears to be unused

tests/cnf/ran/ptp/internal/profiles/nodematcher.go
```diff
@@ -69,6 +78,29 @@ func GetNodeInfoMap(client *clients.Settings) (map[string]*NodeInfo, error) {
...
 }
 
+// buildHardwareConfigIndex lists all HardwareConfig CRs and returns a map keyed by
+// RelatedPtpProfileName. Returns an empty map when the CRD is not installed.
+func buildHardwareConfigIndex(client *clients.Settings) map[string]*ptp.HardwareConfigBuilder {
+	index := make(map[string]*ptp.HardwareConfigBuilder)
+
+	hwConfigs, err := ptp.ListHardwareConfigs(client)
+	if err != nil {
+		klog.V(tsparams.LogLevel).Infof(
```
klaskosk: something AI brought up: we should probably distinguish CRD not found from other errors

`tests/hw-accel/amdgpu/internal/amdgpucommon/errors.go` does this already, although that implementation is a bit redundant. maybe something like this:

```go
import (
	apierrors "k8s.io/apimachinery/pkg/api/errors"
	apimeta "k8s.io/apimachinery/pkg/api/meta"
)

func IsCRDMissing(err error) bool {
	return apierrors.IsNotFound(err) || apimeta.IsNoMatchError(err)
}
```

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -652,8 +651,9 @@ func assertNoFreerunEvent(nodeName string, sinceTime time.Time) {
 
 // changeHoldoverPluginSettings configures holdover thresholds on a PTP profile and waits for the config to
 // take effect. If the desired settings already match the current settings, the function returns immediately.
-// Otherwise, it updates the PtpConfig CR, waits for the daemon to reload profiles, and then waits for the
-// clock to return to LOCKED state with stable metrics. The original settings are restored via DeferCleanup.
+// Otherwise, it updates the holdover settings (via HardwareConfig CR on 4.22+/GNRD, or PtpConfig plugin
+// on pre-4.22), waits for the daemon to reload profiles, and then waits for the clock to return to LOCKED
+// state with stable metrics. The original settings are restored via DeferCleanup.
 func changeHoldoverPluginSettings(
```
klaskosk: nit: `changeHoldoverSettings` may be more appropriate since it's not always plugin settings

tests/cnf/ran/ptp/internal/profiles/profiles.go
```diff
@@ -117,6 +117,9 @@ type ProfileInfo struct {
 	// ConfigIndex is the number in the config file for the ptp4l corresponding to this profile. Profiles should
 	// have a ptp4l process unless they are HA.
 	ConfigIndex *uint
+	// HardwareConfig is the associated HardwareConfig CR for this profile, populated during GetNodeInfoMap.
+	// Nil means the profile uses the PtpConfig plugin path for holdover settings (pre-4.22 or non-GNRD).
+	HardwareConfig *ptp.HardwareConfigBuilder
```
klaskosk, line 122: i'm personally a little conflicted on including the HardwareConfigBuilder in the ProfileInfo since it does break a lot of the expectations of the original struct

that being said, I think the alternative of including an ObjectKey or similar feels like it's being inconvenient solely to abide by rules

we do need to update some of the docs though

> Since profiles can be readily retrieved from the cluster, it only contains information that must be parsed and a reference to the profile on the cluster.

like this doesn't make sense if we include the hardware config builder (which includes an api client and the full hardware config)

tests/cnf/ran/ptp/internal/profiles/profiles.go
```diff
@@ -341,9 +344,10 @@ func parseParentDataSetOutput(output string) (map[string]string, error) {
 // modifications to the cloned ProfileInfo do not affect the original.
 func (profileInfo *ProfileInfo) Clone() *ProfileInfo {
 	clone := &ProfileInfo{
-		ProfileType: profileInfo.ProfileType,
-		Reference:   profileInfo.Reference,
-		Interfaces:  make(map[iface.Name]*InterfaceInfo),
+		ProfileType:    profileInfo.ProfileType,
+		Reference:      profileInfo.Reference,
+		Interfaces:     make(map[iface.Name]*InterfaceInfo),
+		HardwareConfig: profileInfo.HardwareConfig,
```
klaskosk, line 351: same here about docs. a shallow copy of the builder does make sense semantically, but does need documented since it means modifications of the cloned ProfileInfo can have side effects for the original

digression: maybe having a clone function is semantically unnecessary? though interface info is definitely node-specific (clock identities, inter-interface connections, etc.)

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -762,19 +766,70 @@ func discoverHoldoverTestData(
...
 			return true
 		}
 	}
 
 	return false
 }
+
+// getHoldoverSettings reads holdover settings for the profile, dispatching to the HardwareConfig
+// CR path (4.22+/GNRD E825) when one is associated, or the PtpConfig plugin path otherwise.
+func getHoldoverSettings(testData holdoverTestData) (*profiles.HoldoverPluginSettings, error) {
```
klaskosk: GM tests will also need the ability to switch between plugin settings, so this function is a good candidate to be moved to the profiles package

I know we have separate helpers there for the HardwareConfig path and for the plugin path, so it'd be good to also but the combined helpers there. That way future tests can just say "here's my profile, get the holdover settings in the way that makes sense" instead of reimplementing the logic again

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -762,19 +766,70 @@ func discoverHoldoverTestData(
...
+	if err != nil {
+		return nil, err
+	}
+
+	return profiles.GetHoldoverPluginSettings(ptpProfile)
+}
+
+// applyHoldoverSettings writes holdover settings for the profile, dispatching by the same rule
+// as getHoldoverSettings. For the plugin path, it also updates the PtpConfig CR on the cluster.
+func applyHoldoverSettings(testData holdoverTestData, settings profiles.HoldoverPluginSettings) error {
```
klaskosk: same as for the get function, this would go well in the profiles package

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -762,19 +766,70 @@ func discoverHoldoverTestData(
...
+	}
+
+	_, err = ptpConfig.Update()
+
+	return err
+}
+
+// waitForSettingsApplied waits for the daemon to acknowledge a holdover settings change.
+// HardwareConfig changes emit "Successfully applied profile"; PtpConfig changes emit "load profiles".
+func waitForSettingsApplied(testData holdoverTestData, since time.Time, timeout time.Duration) error {
```
klaskosk: same as above, I think this is a good candidate for moving to the profiles package

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -734,26 +722,42 @@ func discoverHoldoverTestData(
...
+
+				upstreamPort = port
+			} else {
+				// pre-4.22 / E810: upstream port from plugin interconnections.
+				// E830 without a HardwareConfig has no supported holdover path — skip it.
+				if hasE830Plugin(ptpProfile) {
+					continue
+				}
+
+				port, portErr := profiles.GetUpstreamPort(ptpProfile)
```
klaskosk: if `GetUpstreamPortForProfile` is the function that handles the logic for deciding where to get the upstream port, then we should stick with that instead of using the lower level function too

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -762,19 +766,70 @@ func discoverHoldoverTestData(
 	return nil
 }
 
-// isUnsupportedPlugin checks whether the profile uses a plugin type that does not support holdover tests
-// (e825 or e830).
-func isUnsupportedPlugin(profile *ptpv1.PtpProfile) bool {
+// hasE830Plugin reports whether the profile contains an E830 plugin entry.
+func hasE830Plugin(profile *ptpv1.PtpProfile) bool {
```
klaskosk: now that this function has moved from a specific check for a specific test to a generic check for containing a plugin, it may be better placed in the profiles/plugins.go file

although if we're back to "does a specific plugin exist" then that's a quick map lookup on the profile so we probably don't need to list all plugin types then iterate

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -734,26 +722,42 @@ func discoverHoldoverTestData(
...
-			ptpConfigBuilder, pullErr := profileInfo.Reference.PullPtpConfig(RANConfig.Spoke1APIClient)
-			Expect(pullErr).ToNot(HaveOccurred())
+					continue
+				}
+
+				upstreamPort = port
+			} else {
+				// pre-4.22 / E810: upstream port from plugin interconnections.
+				// E830 without a HardwareConfig has no supported holdover path — skip it.
+				if hasE830Plugin(ptpProfile) {
```
klaskosk: is e825 without HardwareConfig supported? My understanding was that you could do that to allow the BC w/o Holdover use case

if that's the case, then the change to only exclude E830 in the no hardware config path is incorrect, right?

I trust your expertise here, I just want to make sure we won't run these tests if a no holdover GNR-D config is applied

klaskosk (commented):
went ahead and reviewed early in case it helps shorten the review cycle. understandably a draft still so feel free to disregard anything irrelevant

most comments are naming/docs tweaks and moving helpers to the internal package (since I've just reviewed the GM PR which can use them 🙂)

definitely feeling that the profiles package has some structural issues, but these changes do a good job at working with them so I appreciate you working with what we have for now

tests/cnf/ran/ptp/internal/profiles/nodematcher.go
```diff
@@ -69,6 +80,32 @@ func GetNodeInfoMap(client *clients.Settings) (map[string]*NodeInfo, error) {
...
+// RelatedPtpProfileName. Returns an empty map when the CRD is not installed.
+func buildHardwareConfigIndex(client *clients.Settings) map[string]*ptp.HardwareConfigBuilder {
+	index := make(map[string]*ptp.HardwareConfigBuilder)
+
+	hwConfigs, err := ptp.ListHardwareConfigs(client)
+	if err != nil {
+		if apierrors.IsNotFound(err) || apimeta.IsNoMatchError(err) {
+			klog.V(tsparams.LogLevel).Infof("HardwareConfig CRD not installed (pre-4.22 cluster): %v", err)
+		} else {
+			klog.Warningf("Failed to list HardwareConfigs: %v", err)
```
klaskosk: minor: please use `klog.V(tsparams.LogLevel)`

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -734,26 +721,36 @@ func discoverHoldoverTestData(
...
+
+				upstreamPort = port
+			} else {
+				port, portErr := profiles.GetUpstreamPortForProfile(ptpProfile)
+				if portErr != nil {
+					continue
+				}
+
+				upstreamPort = port
+			}
```
klaskosk: we should be able to simplify this if statement now that the two branches are identical

klaskosk (commented):
appreciate the update! I added two more comments and then there's 3 remaining from the last review

the old ones in `profiles.go` are just a request to update the doc comments and then I'd still like to see the `getHoldoverSettings` from `ptp-holdover.go` moved to a helper

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -731,29 +716,19 @@ func discoverHoldoverTestData(
...
-		matchingProfiles := nodeInfo.GetProfilesByTypes(profileType)
-
-		for _, profileInfo := range matchingProfiles {
-			profile, pullErr := profileInfo.PullProfile(RANConfig.Spoke1APIClient)
+		for _, profileInfo := range nodeInfo.GetProfilesByTypes(profileType) {
+			ptpProfile, pullErr := profileInfo.PullProfile(RANConfig.Spoke1APIClient)
 			Expect(pullErr).ToNot(HaveOccurred())
 
-			if isUnsupportedPlugin(profile) {
+			upstreamPort, ok := holdoverUpstreamPort(profileInfo, ptpProfile, name)
```
klaskosk: I feel like this tree of functions can be simplified by adding a condition like  (pseudocode) `if !HasPlugin(E810) && HardwareConfig == nil { continue }` to the loop. Only E810 is supported without a HardwareConfig. I know you've done the actual testing so let me know if I'm thinking incorrectly here

My original comment here was that we shouldn't test E825 without HardwareConfig (not just E830), but I missed in the second round that the branches being identical was incorrect, so my comment to combine the branches had a faulty basis

tests/cnf/ran/ptp/internal/profiles/profiles.go
klaskosk: @dpopsuev this is the last unresolved thread I see, it'd be good to have the documentation updated here so that it acknowledges the HardwareConfig builder pointer **is** shared and mutable by others who hold it

klaskosk:
Assuming testing works out for this PR, the only outstanding thing is a docs update on the `ProfileInfo.Clone` method. and probably a rebase since go.mod shouldn't need any updates anymore. but other than those minor things the changes lgtm!

klaskosk:
code lgtm now!

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -731,50 +714,35 @@ func discoverHoldoverTestData(
...
-			profile, pullErr := profileInfo.PullProfile(RANConfig.Spoke1APIClient)
+		for _, profileInfo := range nodeInfo.GetProfilesByTypes(profileType) {
+			ptpProfile, pullErr := profileInfo.PullProfile(RANConfig.Spoke1APIClient)
 			Expect(pullErr).ToNot(HaveOccurred())
 
-			if isUnsupportedPlugin(profile) {
+			if profileInfo.HardwareConfig == nil && !profiles.HasPlugin(ptpProfile, ptp.PluginTypeE810) {
+				klog.V(tsparams.LogLevel).Infof(
+					"Skipping profile %s on node %s: unsupported holdover path: %+v",
+					profileInfo.Reference.ProfileName, name, ptpProfile.Plugins)
```
dpopsuev, line 724: @klaskosk I added this last minute, if for whatever reason we skip tests which we don't want to skip then we could have something to grep at.

klaskosk: I think this is a good change, thanks

klaskosk (approved):

PR #1429, opened 2026-06-08
cnf network: fix QinQ tcpdump capture by enabling line-buffered output

evgenLevin:
Without the -l flag, tcpdump uses full buffering when stdout is redirected to a file, causing the capture file to appear empty when read by tail. Also increase tail line count from 20 to 100 to ensure QinQ frames are visible even when broadcast traffic is present.

## Summary by CodeRabbit

* **Tests**
  * Enhanced QinQ traffic test capture with improved line-buffering and expanded data collection for standard and DPDK test clients, increasing test reliability and observability.


ajaggapa (approved):
/lgtm

PR #1448, opened 2026-06-12
CNF-17455   QE Automation: improve basic ptp sync test case

Bonnie-Block:
When checking lock states in metrics, explicitly checks for the presence of metrics expected based on configured ptp profiles.  Explicit checks:

- phc2sys clock state for CLOCK_REALTIME
- ptp4l clock state for each slave interface
- gnss clock state for GM config
- dpll clock state:

DPLL clock state presence is explicitly checked when all of the following are true for a given profile:

DPLL capability exists -- the profile has either:

An Intel plugin (E810/E825/E830) with DpllSettings configured, or
A HardwareConfig CR with DPLL holdover parameters in its ClockChain.
At least one DPLL interface is discoverable -- determined by:

For GM variants (T-GM, Multi-NIC GM, NTP Fallback): plugin-based discovery (E810 RX pins or E825/E830 Devices).
For all other profiles (T-BC, T-TSC, OC, etc.): the profile's client (upstream) interfaces.
Falls back from plugin discovery to client interfaces if the plugin path fails.
The discovered interface has a valid NIC name -- ifaceName.GetNIC() must return a non-empty string, since DPLL metrics are reported at the NIC level.

If all three conditions hold, a process=dpll, iface=<nic> entry is added to the expected clock states list and asserted as LOCKED via EnsureClocksAreLocked.

## Summary by CodeRabbit

* **Tests**
  * Enhanced PTP clock-state metric validation to verify expected metrics are present in monitoring systems before running assertions.
  * Improved error reporting with detailed information about missing expected clock-state entries.
  * Updated test suite to automatically compute and validate expected clock states based on cluster configuration.

### DIFF
--- a/tests/cnf/ran/ptp/internal/metrics/ensurelocked.go
+++ b/tests/cnf/ran/ptp/internal/metrics/ensurelocked.go
@@ -3,17 +3,58 @@ package metrics
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/tsparams"
+	"k8s.io/klog/v2"
 )
+
+// ensureLockedConfig holds the configuration for EnsureClocksAreLocked and EnsureClocksAreStable.
+type ensureLockedConfig struct {
+	expectedClockStates []ExpectedClockState
+}
+
+// EnsureLockedOption configures optional behavior for EnsureClocksAreLocked and EnsureClocksAreStable.
+type EnsureLockedOption func(*ensureLockedConfig)
+
+// WithExpectedClockStates configures the presence check phase. When provided, EnsureClocksAreLocked will first
+// verify that all expected clock state metrics are present (at least one sample returned per entry) before
+// proceeding to the locked assertion. This ensures that missing metrics (e.g., a dpll or gnss process that
+// never started) are caught rather than silently passing.
+func WithExpectedClockStates(expected []ExpectedClockState) EnsureLockedOption {
+	return func(cfg *ensureLockedConfig) {
+		cfg.expectedClockStates = expected
+	}
+}
 
 // EnsureClocksAreLocked ensures that all PTP clocks are locked across all nodes covered by the Prometheus API client.
 // It is designed to be used as a BeforeEach/AfterEach check to ensure the cluster is in a stable state.
 //
 // It ensures that clocks are locked for 10 seconds with a timeout of 5 minutes. It does not check the clock state of
 // the chronyd process, as it will be FREERUN when PTP is working correctly.
-func EnsureClocksAreLocked(prometheusAPI prometheusv1.API) error {
+//
+// When WithExpectedClockStates is provided, the function first verifies that all expected metric series are present
+// before asserting the locked state.
+func EnsureClocksAreLocked(prometheusAPI prometheusv1.API, opts ...EnsureLockedOption) error {
+	cfg := &ensureLockedConfig{}
+	for _, opt := range opts {
+		opt(cfg)
+	}
+
+	if len(cfg.expectedClockStates) > 0 {
+		deduped := deduplicateExpectedClockStates(cfg.expectedClockStates)
+
+		klog.V(tsparams.LogLevel).Infof("Ensuring expected clock state metrics are present:\n%s",
+			FormatExpectedClockStates(deduped))
+
+		err := ensureClockStatesPresent(prometheusAPI, deduped, 5*time.Minute)
+		if err != nil {
+			return fmt.Errorf("failed presence check before asserting locked state: %w", err)
+		}
+	}
+
 	query := ClockStateQuery{
 		Process: DoesNotEqual(ProcessChronyd),
 	}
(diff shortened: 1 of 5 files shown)


tests/cnf/ran/ptp/internal/profiles/expectedmetrics.go
```diff
@@ -0,0 +1,238 @@
...
+
+// GetExpectedClockStates derives the expected openshift_ptp_clock_state metric series from the PtpConfig
+// configuration on the cluster. It inspects each node's recommended profiles to determine which (process, interface)
+// combinations should be reporting clock state metrics.
+//
+// The expected metrics are:
+//   - process=phc2sys, iface=CLOCK_REALTIME: for each profile that runs phc2sys (all non-TBCTransmitter profiles)
+//   - process=ptp4l, iface=<interface>: for each slave (client) interface. Simple OC profiles use raw names
+//     (e.g., "ens1f0"); T-BC receiver profiles use NIC names (e.g., "enox") because boundary_clock_jbod mode
+//     consolidates the ptp4l clock state to the NIC level.
```
klaskosk, line 20: I think all profile types use the NIC name/alias for `openshift_ptp_clock_state`. looking at a node with a simple OC config, I see `iface="ens1fx"` for a profiles with `ens1f1np1`

I'm not sure if there's an easy way to check which it'll be 😕. since it seems related to whether there's multiple profiles for the same NIC

tests/cnf/ran/ptp/internal/profiles/plugins.go
```diff
@@ -289,6 +289,36 @@ func GetRxInterfaces(profile *ptpv1.PtpProfile) ([]iface.Name, error) {
 	return getInterfacesWithPluginPins(profile, ptp.PluginTypeE810, PinStateRx)
 }
 
+// GetDpllInterfaces returns the interfaces that have DPLL hardware monitoring enabled.
+// The discovery method depends on the Intel plugin type present in the profile:
+//   - E810: interfaces with RX pins (pin state 1), since DPLL monitors the receive path.
```
klaskosk, line 294: dpll shows up on all interfaces of a multi-NIC WPC profile from what I see. I think it's still there on single NIC too?

tests/cnf/ran/ptp/internal/metrics/ensurelocked.go
```diff
@@ -3,17 +3,58 @@ package metrics
...
 // EnsureClocksAreLocked ensures that all PTP clocks are locked across all nodes covered by the Prometheus API client.
 // It is designed to be used as a BeforeEach/AfterEach check to ensure the cluster is in a stable state.
 //
 // It ensures that clocks are locked for 10 seconds with a timeout of 5 minutes. It does not check the clock state of
 // the chronyd process, as it will be FREERUN when PTP is working correctly.
-func EnsureClocksAreLocked(prometheusAPI prometheusv1.API) error {
+//
+// When WithExpectedClockStates is provided, the function first verifies that all expected metric series are present
+// before asserting the locked state.
+func EnsureClocksAreLocked(prometheusAPI prometheusv1.API, opts ...EnsureLockedOption) error {
```
klaskosk, line 40: for integrating the existence checks, let's create a separate function for those. The EnsureClocksAreLocked/EnsureClocksAreStable can be quick checks to make sure nothing's in freerun and then the tests specifically for sync can do the more extensive checks for existence and locking

tests/cnf/ran/ptp/internal/profiles/expectedmetrics.go
```diff
@@ -0,0 +1,240 @@
...
+
+	return expected
+}
+
+// isGMProfile returns true if the profile type is a grandmaster variant that may have GNSS hardware.
+func isGMProfile(profileType PtpProfileType) bool {
+	switch profileType {
+	case ProfileTypeGM, ProfileTypeMultiNICGM, ProfileTypeNTPFallback:
+		return true
+	case ProfileTypeOC, ProfileTypeTwoPortOC, ProfileTypeBC, ProfileTypeHA,
```
klaskosk, line 109: minor: adding a default false will be easier, especially since we'll be adding more profile types in the future

tests/cnf/ran/ptp/internal/profiles/expectedmetrics.go
```diff
@@ -0,0 +1,240 @@
...
+
+// ptp4lUsesNICName returns true if the profile type reports ptp4l clock state metrics with NIC names rather
+// than raw interface names. When ptp4l manages multiple ports (clock_type BC, or boundary_clock_jbod mode),
+// the PTP operator consolidates the ptp4l clock state to the NIC level (e.g., "ens3fx" instead of "ens3f2").
+// Single-port OC profiles report per-port with raw interface names.
+func ptp4lUsesNICName(profileType PtpProfileType) bool {
+	switch profileType {
+	case ProfileTypeBC, ProfileTypeTBCReceiver:
+		return true
+	case ProfileTypeOC, ProfileTypeTwoPortOC, ProfileTypeHA, ProfileTypeGM,
```
klaskosk, line 125: minor: see above w.r.t. default false rather than listing

tests/cnf/ran/ptp/internal/profiles/expectedmetrics.go
```diff
@@ -0,0 +1,240 @@
...
+	// do not sync a local clock via phc2sys).
+	if profileInfo.ProfileType != ProfileTypeTBCTransmitter {
+		expected = append(expected, metrics.ExpectedClockState{
+			Process:   metrics.ProcessPHC2SYS,
+			Interface: string(iface.ClockRealtime),
+			Node:      nodeName,
+		})
+	}
+
+	// ptp4l: expected for each slave (client) interface. The interface name format depends on the profile
```
klaskosk, line 60: See [my thread with David EDC](https://redhat-internal.slack.com/archives/C07PT2A0DRA/p1782320113947809) for some more context here

Here's the AI answer on when the operator uses the name directly:

```text
  openshift_ptp_clock_state uses the alias (e.g. ens1fx) when all ports on a PHC share one alias and no other PHC would collide with that alias.

  It uses the full interface name (e.g. ens1f0, ens1f1) when:

  1. Multiple different PHCs would map to the same alias (your CX-6 per-port PHC case), or
  2. Multiple ports on the same PHC don't agree on alias (mixed naming), or
  3. The interface name doesn't match the Intel/Mellanox regex (falls back to the raw name).
```

I think we can focus on the first case to simplify our tests. Though the logic there is still pretty complex: if we have multiple single-port OC profiles which each map to the same iface.NICName (alias), then we need to use the iface.Name directly

PR #1449, opened 2026-06-14
cnf network: fix QinQ tcpdump capture by enabling line-buffered outpu…

gkopels:
…t (#1429)

Without the -l flag, tcpdump uses full buffering when stdout is redirected to a file, causing the capture file to appear empty when read by tail. Also increase tail line count from 20 to 100 to ensure QinQ frames are visible even when broadcast traffic is present.

Cherry pick backport from 4.22


evgenLevin (approved):
lgtm

PR #1452, opened 2026-06-15
[PTP] Add T-GM ptpconfig lifecycle regression test for OCPBUGS-58131

hhassid:
## Summary

Adds a regression test for OCPBUGS-58131 / CNF-20352 that validates
the `/gpsd/data` file does not accumulate data after repeated T-GM
PtpConfig delete and re-apply cycles.

Jira: [CNF-20352 - Corrective Measure for OCPBUGS-58131: [4.16.42] PTP State does not recover after updating the GM ptpconfig](https://redhat.atlassian.net/browse/CNF-20352)
Bug:  [OCPBUGS-58131 - [4.16.42] PTP State does not recover after updating the GM ptpconfig](https://redhat.atlassian.net/browse/OCPBUGS-58131)

## Key Changes
- Added Regression Validation: Introduced validateGpsdFileEmpty to assert that /gpsd/data is zero-size or absent. It uses a Consistently block to poll the file state every second over a 5-second window, failing immediately if the file accumulates data.
- Integrated with GNSS Loss Tests: Placed /gpsd/data validations inside the ptp-gnss-loss.go test cases directly after PtpConfig changes and clock stabilization.
- Removed Redundant Suite: Dropped the initially proposed ptp-gm-ptpconfig-lifecycle.go (which ran 10 forced delete/apply cycles) in favor of the more efficient, integrated approach.
- Teardown Optimization: Added an early return to the AfterEach block in ptp-gnss-loss.go to gracefully skip PtpConfig restoration and teardown procedures if the test spec was skipped.

## Test plan

- [x] Run the PTP T-GM GNSS Loss test suite.
  
  Co-authored-by: Cursor <cursoragent@cursor.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GNSS-loss test coverage to skip cleanup checks after skipped scenarios, avoiding false failures.
  * Added an extra validation step to confirm GPS data remains empty during key PTP daemon transitions.
  * Strengthened three GNSS-loss transition checks to better catch regressions after configuration changes and clock stabilization.


hhassid:
Not tested yet but ready for first review
@klaskosk @dpopsuev @Bonnie-Block

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
```diff
@@ -0,0 +1,253 @@
...
+
+		By("ensuring clocks are locked after testing")
+
+		err = metrics.EnsureClocksAreLocked(prometheusAPI)
+		Expect(err).ToNot(HaveOccurred(), "Failed to assert clock state is locked")
+	})
+
+	// CNF-20352 - validate /gpsd/data file does not grow after repeated T-GM ptpconfig delete/apply cycles
+	It("validates /gpsd/data does not grow after repeated GM ptpconfig delete and apply cycles",
+		reportxml.ID("CNF-20352"), func() {
```
klaskosk: please add this test case to polarion and use that ID. you should be able to create the workitem without actually importing a plan from what I remember

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
```diff
@@ -0,0 +1,253 @@
...
+			// Track configs that have already been cycled. When multiple GM nodes share
+			// the same PtpConfig, cycling it once exercises all of them — running
+			// additional cycles for the same config would be redundant.
+			cycledConfigs := make(map[string]bool)
+
+			nodeInfoMap, err := profiles.GetNodeInfoMap(RANConfig.Spoke1APIClient)
+			Expect(err).ToNot(HaveOccurred(), "Failed to get node info map")
+
+			for _, nodeInfo := range nodeInfoMap {
+				if nodeInfo.Counts[profiles.ProfileTypeGM] == 0 &&
```
klaskosk: I think this would apply to `ProfileTypeNTPFallback` too. unless you get into holdover/freerun conditions, `ntpfailover` behaves like an ordinary GM profile (mostly)

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
```diff
@@ -0,0 +1,253 @@
...
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/ptpdaemon"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/tsparams"
+)
+
+const (
+	gpsdDataPath         = "/gpsd/data"
+	validateAttempts     = 5
+	validateInterval     = 1 * time.Second
+	profileReloadTimeout = 5 * time.Minute
+	lifecycleCycles      = 10
```
klaskosk: I'm curious how long this test takes to run with this many cycles and if we could get away with fewer

based on the bug saying it occurs 60% of the time, we have a 1% chance of not seeing it in 5 cycles vs 0.01% chance in 10 cycles

may be an opportunity to optimize a little given the time constraint in the GM lane

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
```diff
@@ -0,0 +1,253 @@
...
+)
+
+var _ = Describe("PTP T-GM PtpConfig Lifecycle", Label(tsparams.LabelGMPtpConfigLifecycle), func() {
+	var (
+		prometheusAPI   prometheusv1.API
+		savedPtpConfigs []*ptp.PtpConfigBuilder
+	)
+
+	BeforeEach(func() {
```
klaskosk: maybe a version check is necessary? I'm not sure how far the fix got backported

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
```diff
@@ -0,0 +1,253 @@
...
+					By(fmt.Sprintf("%s deleting GM ptpconfig %s on node %s",
+						cycleLabel, gmConfigName, nodeInfo.Name))
+
+					deleteTime := time.Now()
+
+					err = gmSavedConfig.Delete()
+					Expect(err).ToNot(HaveOccurred(),
+						"Failed to delete GM ptpconfig on cycle %d on node %s", cycle+1, nodeInfo.Name)
+
+					// The daemon logs "load profiles" whenever it reloads its configuration.
```
klaskosk: one thing I've found with GM profiles is that they take a relatively long time to load. the best check I've found so far is waiting till ts2phc is locked. based on my read of the bug, that may be important since the process needs to start before the file gets filled with data

klaskosk (commented):
overall looks good, a few minor things and some questions to answer from testing

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
hhassid: Thanks for the reminder. Added.

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
hhassid: Added.

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
hhassid: ts2phc validation should be added after a configuration is applied. In this step, after the removing of the configuration, ts2phc will not run.

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
hhassid: I thought about it. I used to run this using bash script and it took les then an hour but not sure how long it will be using ecogo.

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
hhassid: Added check to run since 4.16

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
klaskosk: oh great point about the removal, thanks!

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
hhassid: @klaskosk 
https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/Private/job/hhassid/job/ocp-far-edge-vran-tests/38/
This one took 42min

klaskosk: thanks! I think Yang's suggestion of moving to 1 check makes sense, we'd have ~94% chance of seeing once per sprint based on the 60% number.

I know she mentioned possibly adding the check to other cases where we update the config. I think that'd work too to verify the bug since I think the operator treats updates the same as new configs in terms of starting up all the processes

I personally prefer the idea of using a single test case a little more since it keeps the check explicit and prevents it being unrelated to the other tests. although I'm not opposed to including the check in other case(s) instead if you think that's a better approach

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
```diff
@@ -0,0 +1,271 @@
...
+		Expect(err).ToNot(HaveOccurred(), "Failed to assert clock state is locked")
+
+		By("saving PtpConfigs before testing")
+
+		savedPtpConfigs, err = profiles.SavePtpConfigs(RANConfig.Spoke1APIClient)
+		Expect(err).ToNot(HaveOccurred(), "Failed to save PtpConfigs")
+	})
+
+	AfterEach(func() {
+		By("restoring PtpConfigs after testing")
```
klaskosk: minor: adding a check for if the case was skipped would save some unnecessary checks

```go
if CurrentSpecReport().State == types.SpecStateSkipped {
	return
}
```

tests/cnf/ran/ptp/tests/ptp-gm-ptpconfig-lifecycle.go
hhassid: Although I think we should keep the test cases sparely, in this case it will be more reasonable to add a validation to T-GM (not sure regarding ntp-fallback) tests so we can monitor the file in multiple cases and still not increase much the running time.

I'll remove the file and add the validation function to gnss-loss file.

klaskosk: Sounds good! please mark the test case as inactive in Polarion then (and add a short note explaining why so we're not confused like with 70111 :slightly_smiling_face:)

also, please add a comment to the validation function that this is a corrective measure for OCPBUGS-58131. It'll be nice to preserve why the code exists for the future so we don't delete in a refactor

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -155,6 +161,10 @@ var _ = Describe("PTP T-GM GNSS Loss", Label(tsparams.LabelGNSSLoss), func() {
 
 					err = metrics.EnsureClocksAreStable(prometheusAPI, 1*time.Minute)
 					Expect(err).ToNot(HaveOccurred(), "Failed to ensure clocks are stable after config change")
+
+					By("validating /gpsd/data is not growing after config change on node " + nodeName)
+
+					validateGpsdFileEmpty(nodeName)
```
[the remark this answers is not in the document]
hhassid: ptpConfig.Update() should remove and create /gpsd/data so it acts the same as delete/re-apply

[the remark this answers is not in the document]
dpopsuev: Yes, Update is a masked Delete & Apply.

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -714,3 +732,26 @@ var _ = Describe("PTP T-GM GNSS Loss", Label(tsparams.LabelGNSSLoss), func() {
...
+	const (
+		validateAttempts = 5
+		validateInterval = 1 * time.Second
+	)
+
+	Consistently(func() error {
+		_, err := ptpdaemon.ExecuteCommandInPtpDaemonPod(
+			RANConfig.Spoke1APIClient,
+			nodeName,
+			fmt.Sprintf("[ ! -s %s ]", "/gpsd/data"),
```
dpopsuev: @hhassid nit: redundant usage of Sprintf; you know the path [/gpsd/data] in compile time, reserve using the Sprint* methods for runtime.

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/reportxml"
 	"k8s.io/klog/v2"
 
+	"github.com/onsi/ginkgo/v2/types"
```
dpopsuev: @hhassid nit: the import clustering / group is usually:

```
import {
  "remote" [outside the code base]

  "local" [within the code base]
}
```

dpopsuev (commented):
Good commit, well scoped!
Comments are pure nits about 'Idiomatic Go'.

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
hhassid: Thanks. I started to change this line and didn't finish...
Fixed

dpopsuev (approved):

klaskosk (approved):
thanks! and thank you Daniel for reviewing!

PR #1466, opened 2026-06-18
[PTP] Migrate GM NMEA loss test 70111 from cnf-gotests

dpopsuev:
## Summary
- Migrate Polarion test case 70111 (GM NMEA loss / recovery) from cnf-gotests to eco-gotests
- Extract `assertClockClassInMetrics` helper, deduplicating the clock class metric assertion block across tests 78463, 78464, 78465, and 70111
- Fix pre-existing bug in test 78463 where `ClockStateQuery` was not scoped to the node under test

## Details

**Test 70111** verifies the GM clock event sequence when the upstream GNSS sync source is lost and restored:
1. Check NMEA status metric is available (pre-loss baseline)
2. Configure holdover plugin settings for extended holdover
3. Simulate GNSS loss via u-blox `ubxtool`
4. Wait for events: `FAILURE-NOFIX` → clock class 7 → `HOLDOVER`
5. Restore GNSS sync
6. Wait for recovery events: `SYNCHRONIZED` → clock class 6 → `LOCKED`
7. Verify NMEA status metric is available again (post-recovery)
8. Verify clock class 6 in Prometheus metrics

**DRY refactoring**: The clock-class-in-metrics assertion block was copy-pasted across all 4 tests. Extracted into `assertClockClassInMetrics()` helper with `GinkgoHelper()` for proper stack traces.

**Bug fix**: `ClockStateQuery` in test 78463 was missing `Node` filter, causing it to assert clock state across ALL PTP nodes. Added `Node: metrics.Equals(nodeName)` to scope the query.

## Jira
[CNF-20472](https://redhat.atlassian.net/browse/CNF-20472)

## Test plan
- [x] Run `gnss-loss` label on a GM-capable cluster: `ginkgo -L "gnss-loss"`
- [x] Verify 70111 passes on single-NIC and multi-NIC GM configurations
- [x] Verify existing tests 78463, 78464, 78465 still pass with extracted helper
- [x] Verify 78463 clock state assertion works correctly on multi-node clusters

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Enhanced GNSS loss/recovery test coverage by asserting NMEA status availability immediately before simulated GNSS loss and after recovery.
  * Consolidated repeated “clock class 6 in metrics” verification into a reusable helper to reduce duplication and improve maintainability.

### DIFF
--- a/tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
+++ b/tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
@@ -166,6 +166,10 @@ var _ = Describe("PTP T-GM GNSS Loss", Label(tsparams.LabelGNSSLoss), func() {
 					validateGpsdFileEmpty(nodeName)
 				}
 
+				By("checking NMEA status metric is available before GNSS loss on node " + nodeName)
+
+				assertNMEAStatusAvailable(prometheusAPI, nodeName)
+
 				By("simulating GNSS loss on node " + nodeName)
 
 				DeferCleanup(cleanupGNSSSync(prometheusAPI, nodeName, protocolVersion))
@@ -239,28 +243,19 @@ var _ = Describe("PTP T-GM GNSS Loss", Label(tsparams.LabelGNSSLoss), func() {
 				))
 				Expect(err).ToNot(HaveOccurred(), "Failed to receive clock class 6 event on node %s", nodeName)
 
+				By("verifying NMEA status is available after recovery on node " + nodeName)
+
+				assertNMEAStatusAvailable(prometheusAPI, nodeName)
+
 				By("verifying clock class 6 in metrics")
 
-				var configFile string
-				if configSupported {
-					configFile, err = processes.GetPtp4lConfigByRelatedProcess(
-						RANConfig.Spoke1APIClient, nodeName, processes.Ts2phc)
-					Expect(err).ToNot(HaveOccurred(), "Failed to determine ptp4l config for node %s", nodeName)
-				}
-
-				clockClassQuery := metrics.ClockClassQuery{
-					Process: metrics.Equals(metrics.ProcessPTP4L),
-					Node:    metrics.Equals(nodeName),
-					Config:  metrics.Equals(configFile),
-				}
-				err = metrics.AssertQuery(context.TODO(), prometheusAPI, clockClassQuery,
-					metrics.ClockClass6, metrics.AssertWithTimeout(1*time.Minute))
-				Expect(err).ToNot(HaveOccurred(), "Failed to assert clock class is 6 in metrics on node %s", nodeName)
+				assertClockClass6InMetrics(prometheusAPI, nodeName, configSupported)
 
 				By("verifying clock state LOCKED in metrics")
 
 				clockStateQuery := metrics.ClockStateQuery{
 					Process: metrics.DoesNotEqual(metrics.ProcessChronyd),
+					Node:    metrics.Equals(nodeName),
 				}
 				err = metrics.AssertQuery(context.TODO(), prometheusAPI, clockStateQuery,
 					metrics.ClockStateLocked, metrics.AssertWithTimeout(1*time.Minute))
@@ -355,6 +350,10 @@ var _ = Describe("PTP T-GM GNSS Loss", Label(tsparams.LabelGNSSLoss), func() {
 
 					validateGpsdFileEmpty(nodeName)
 				}
+
+				By("checking NMEA status metric is available before GNSS loss on node " + nodeName)
+
+				assertNMEAStatusAvailable(prometheusAPI, nodeName)
 
 				By("simulating GNSS loss on node " + nodeName)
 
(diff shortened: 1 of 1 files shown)


dpopsuev:
@klaskosk Test case 70111 migrated and validated on helix65 (OCP 4.22.1, E810-C). All 4 `gnss-loss` specs pass — the new test plus the 3 existing holdover tests with the refactored clock-class helper. Ready for review.

tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -691,26 +663,210 @@ var _ = Describe("PTP T-GM GNSS Loss", Label(tsparams.LabelGNSSLoss), func() {
...
+
+				ptpConfig, err := gmProfileInfo.Reference.PullPtpConfig(RANConfig.Spoke1APIClient)
+				Expect(err).ToNot(HaveOccurred(), "Failed to pull PtpConfig for node %s", nodeName)
+
+				profileToUpdate := &ptpConfig.Definition.Spec.Profile[gmProfileInfo.Reference.ProfileIndex]
+
+				desiredSettings := profiles.HoldoverPluginSettings{
+					LocalHoldoverTimeout:   14400,
+					LocalMaxHoldoverOffSet: 1500,
+					MaxInSpecOffset:        1000,
```
hhassid: In cnf-gotest there is no config changes.
Those values are the same as in [78463](https://github.com/rh-ecosystem-edge/eco-gotests/blob/main/tests/cnf/ran/ptp/tests/ptp-gnss-loss.go#L127-L130)
I think maybe we can add the NMAE validation in 78463 test in case this is still support 4.18+.

hhassid (commented):
1. Really liked the assert block, very elegant.
2. 70111 test used to run on version [under 4.18](https://gitlab.cee.redhat.com/cnf/cnf-gotests/-/blob/master/test/ran/ptp/tests/ptp_recovery.go?ref_type=heads#L885-887) (Not sure why). All tests under this label run on 4.18 and above
3. This test seems to complete [78463](https://github.com/rh-ecosystem-edge/eco-gotests/blob/main/tests/cnf/ran/ptp/tests/ptp-gnss-loss.go#L96) in case there is need to change the plugin values (no changes in cnf-gotests). The configuration modification takes time and we might think about merge those two.

klaskosk:
+1 to Hen's comments, that answers why the case was marked as inactive in Polarion -- it's not being run anymore

dpopsuev:
@klaskosk @hhassid
So @hhassid suggestion is to collapse the NMEA test into the grander 'sync loss' test due to the overlaps.
I'm for that, but I'm also for making an inventory of all of the test cases and assessing how can we collapse the test suite into a smaller surface; Let's finish the migration verbatim, and then let's make this part of the refactor since I've reserved a ticket for this 'inventory & consolidate' effort.

klaskosk:
@dpopsuev you wrote the original 78463 so I'm curious if you remember the reason to make it a separate test

It seems like these tests could easily be combined then run in all versions. the actual assertions are the same around the state transitions, and checking nmea status would apply even in 4.18+

Unless you know of a reason not to, I'd support just adding the nmea checks/assertion refactoring in this PR and not including a separate 70111. Just need to be sure Polarion reflects the decision

dpopsuev:
@klaskosk The answer and honest truth is "I didn't know any better"; It was my first commit to the code base and my domain understanding was quite shallow.

It's more about housekeeping in Polarion, essentially it will require doing jolly paperwork on that wonderful system! :smiling_face_with_tear: 
If you consider it worthy I'll perform the collapse, the code can commit and Polarion be updated ad-hoc.

klaskosk:
@dpopsuev ah fair enough :slightly_smiling_face:. Just wanted to be sure there's nothing subtle I'm missing since I wasn't involved with PTP at the time

But sounds like a good plan! I'm glad Hen pointed out the similarity!

dpopsuev:
Folded test 70111 into the three existing gnss-loss tests (78463, 78464, 78465) per the discussion above. The separate NMEA loss spec is removed — instead, an `assertNMEAStatusAvailable` helper now checks NMEA metric availability before GNSS loss and after recovery in every test that triggers clock loss.

dpopsuev:
Updated the code based on your reviews @hhassid @klaskosk, folded the NMEA checks into all three existing gnss-loss tests and dropped the separate 70111 spec. Going to re-run the suite on helix65 for sanity, will update once results are in.

Regarding Polarion, I'll file a ticket to track the housekeeping (deactivating 70111 and updating test case mappings to reflect the consolidation).

dpopsuev:
Tests validated on helix65 (OCP 5.0, PTP 5.0, E810-C). All 3 gnss-loss specs pass with the folded NMEA checks. Suite completed in ~20 minutes.

@klaskosk @hhassid ready for review.

klaskosk (dismissed):

klaskosk (approved):

PR #1477, opened 2026-06-24
ocp network: configure switch trunk mode for 802.1Q QinQ cross-node tests

evgenLevin:
The 802.1Q cross-node test (71679) failed because switch ports were never configured for trunk mode when running with non-E810 NICs. The trunk configuration was only applied inside the 802.1AD context, which is skipped for Mellanox NICs. Add enableQinQTrunkOnSwitch to the 802.1Q BeforeAll so cross-node tagged traffic flows regardless of NIC vendor.

## Summary by CodeRabbit

* **Tests**
  * Improved QinQ test setup to configure switch ports for VLAN trunking before running 802.1Q scenarios.
  * Added stronger validation around switch configuration so related test cases can fail fast if setup does not succeed.

### DIFF
--- a/tests/ocp/sriov/tests/qinq.go
+++ b/tests/ocp/sriov/tests/qinq.go
@@ -411,6 +411,11 @@ var _ = Describe(
 
 		Context("802.1Q", func() {
 			BeforeAll(func() {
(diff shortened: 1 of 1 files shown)


ajaggapa (approved):
/lgtm

PR #1492, opened 2026-07-01
ptp: add CLOCK_REALTIME FREERUN assertion to T-BC holdover freerun tests

hhassid:
Adds the os-clock-sync-state FREERUN event assertion for CLOCK_REALTIME to the T-BC holdover freerun tests (83298, 83300), matching the behavior already validated in the T-GM tests (78464, 78465).

The assertion is added inside assertFreerunState, which is shared by:

83298 – T-BC transition from holdover-in-spec to freerun when LocalMaxHoldoverOffset is reached
83300 – T-BC transition from holdover-out-of-spec to freerun when LocalMaxHoldoverOffset is reached
The new step waits for an OsClockSyncStateChange event with state FREERUN on interface CLOCK_REALTIME, placed after the PTP state/clock-class event checks and before Prometheus metrics validation — consistent with the ordering in the T-GM tests.

### Jira
[CNF-23735](https://redhat.atlassian.net/browse/CNF-23735)

### Test plan

- [x]  Run 83298 83300  tests and verify the CLOCK_REALTIME FREERUN assertion passes

### Key Changes
- Version Gating: Added a PTP operator version check (>= 4.20.0-0) to determine if the os-clock-sync-state FREERUN event should be asserted.
- Event Validation: Introduced logic in ptp-gnss-loss.go and ptp-holdover.go to explicitly wait for the os-clock-sync-state to transition to FREERUN for CLOCK_REALTIME.
- Helper Updates: Updated assertFreerunState, assertHoldoverInSpecToFreerun, and assertHoldoverOutOfSpecToFreerun testing helpers to conditionally run this new validation based on the version check.

Assisted-by: Cursor

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of PTP holdover and GNSS-loss behavior.
  * Added checks for operating system clock FREERUN events when supported.
  * Updated scenarios to distinguish when FREERUN events are expected.

### DIFF
--- a/tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
+++ b/tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
@@ -31,9 +31,10 @@ import (
 
 var _ = Describe("PTP T-GM GNSS Loss", Label(tsparams.LabelGNSSLoss), func() {
 	var (
-		prometheusAPI   prometheusv1.API
-		savedPtpConfigs []*ptp.PtpConfigBuilder
-		configSupported bool
+		prometheusAPI        prometheusv1.API
+		savedPtpConfigs      []*ptp.PtpConfigBuilder
+		configSupported      bool
+		assertOsClockFreerun bool
 	)
 
 	eventTimeout := 5 * time.Minute
@@ -67,6 +68,9 @@ var _ = Describe("PTP T-GM GNSS Loss", Label(tsparams.LabelGNSSLoss), func() {
 		configSupported, err = version.IsVersionStringInRange(
 			RANConfig.Spoke1OperatorVersions[ranparam.PTP], "4.20.0-0", "")
 		Expect(err).ToNot(HaveOccurred(), "Failed to check PTP config label version range")
+
+		// OsClockSyncStateChange FREERUN event is supported from PTP version 4.20.0
+		assertOsClockFreerun = configSupported
 	})
 
 	AfterEach(func() {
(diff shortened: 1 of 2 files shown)


tests/cnf/ran/ptp/tests/ptp-gnss-loss.go
```diff
@@ -67,6 +68,10 @@ var _ = Describe("PTP T-GM GNSS Loss", Label(tsparams.LabelGNSSLoss), func() {
 		configSupported, err = version.IsVersionStringInRange(
 			RANConfig.Spoke1OperatorVersions[ranparam.PTP], "4.20.0-0", "")
 		Expect(err).ToNot(HaveOccurred(), "Failed to check PTP config label version range")
+
+		assertOsClockFreerun, err = version.IsVersionStringInRange(
+			RANConfig.Spoke1OperatorVersions[ranparam.PTP], "4.20.0-0", "")
+		Expect(err).ToNot(HaveOccurred(), "Failed to check PTP version range for CLOCK_REALTIME FREERUN event")
```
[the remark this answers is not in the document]
klaskosk: maybe we should treat the both bounds empty case as a failure... but out of scope for this PR

klaskosk (dismissed):
lgtm, it'd be good to have @dpopsuev confirm the `ptp-holdover.go` changes though

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -199,7 +199,7 @@ var _ = Describe("PTP Holdover", Label(tsparams.LabelTBCTSCHoldover), func() {
 		It("verifies t-tsc transition from holdover-in-spec to freerun when localmaxholdoveroffset reached",
 			reportxml.ID("88275"), func() {
 				assertHoldoverInSpecToFreerun(testData, holdoverPluginSettingsNoOutOfSpec,
-					timeout, expectedClockClasses, clockClassChanges)
+					timeout, expectedClockClasses, clockClassChanges, false)
```
dpopsuev: @hhassid why not assert on a T-TSC?
Unless I'm missing something it should also emit those identical events, meaning we can avoid the `if` statement and the signature update.

hhassid: As in this bug https://redhat.atlassian.net/browse/OCPBUGS-62812 the CLOCK-REALTIME is updated for T-BC and T-BC.
I don't know if there is a requirement for T-TSC, it might be with the same behavior but didn't checked it.

hhassid: @dpopsuev
Removed `assertOsClockFreerun` flag

dpopsuev (approved):
/LGTM

klaskosk (approved):

PR #1498, opened 2026-07-07
cnf network: fix bgp route check [4.18]

evgenLevin:
Backport of PR #950 from release-4.19 to fix a route advertisement timing race in validatePrefix. With frr-k8s v0.0.16, route advertisement takes 2-5s longer after config reload. The previous code checked routes synchronously without retry, causing intermittent failures.

The fix wraps the route presence check in an Eventually loop that polls for the specific subnet until it appears (up to 1 minute). Also adds configurable prefix length parameter and noRouteCheck option from 4.19.


ajaggapa (approved):
/lgtm

PR #1502, opened 2026-07-09
cnf network: wait for NMState policy availability before using interfaces

evgenLevin:
enableEthernetInterfaceWithIP and createSecondaryInterfaceOnNode were fire-and-forget: they created NMState policies but never waited for them to be applied. NMState policies take ~2.5-3 minutes to apply, but subsequent test steps expected the interfaces to be ready within 30-60 seconds.

Use netnmstate.CreatePolicyAndWaitUntilItsAvailable to block until the policy reaches Available state. Also increase the getLinkLocalAddress timeout from 30s to 5m and the static route retry from 1m to 5m as safety margins.

Fixes tests 80393 (BGP Unnumbered) and 75248 (iBGP multihop over secondary interface).


ajaggapa (approved):
/lgtm

PR #1521, opened 2026-07-19
cnf ran ptp: fix event API version check

klaskosk:
This commit fixes an issue where 4.15 and earlier PTP versions would fail to run tests because the API version in the event config does not exist in the CR. The fix defaults to v1 for versions less than 4.16.

Assisted-by: Cursor

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event API version selection across supported PTP versions.
  * Added validation and clearer error handling for unsupported API version settings.
  * Preserved automatic compatibility behavior for older and newer PTP releases.

* **Tests**
  * Added coverage for supported versions, defaults, and invalid configuration values.
  * Included the consumer package in the automated Go unit test run.


klaskosk (commented):
Needs tested on a cluster still.

klaskosk:
Going to merge this now since the 4.14 lanes will be blocked if not, so we really can't make it worse by merging

PR #1542, opened 2026-07-28
cnf ran,system-tests: update O-RAN tests for new URL validation

klaskosk:
This commit switches the existing O-RAN tests for both FT and ST to use the mock-smo for subscriber callbacks rather than a separate oran-subscriber app. The new oran-mock-smo package is designed to be a very similar replacement for the old oran-subscriber package.

This switch is motivated by changes to the o-cloud-manager which now validates that subscription callbacks are the same domain as the SMO. If they are not, the service returns a 400 Bad Request.

To get around this, the image of mock-smo used by FT and ST has been updated with a special endpoint which echoes the callbacks to the pod logs.

Assisted-by: Cursor


klaskosk:
Tested the FT tests for alarms: https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-far-edge-vran-tests/7161

It used to be that 4 of these would fail due to hitting `400 Bad Request` responses

klaskosk:
Latest changes are just a markdown update to document the env vars so don't affect test results

Demostenes777 (approved):

josclark42 (approved):

PR #1557, opened 2026-08-04
[4.21]network: fix unnumbered test cleanup

ajaggapa:
Backport https://github.com/rh-ecosystem-edge/eco-gotests/pull/1559

Fixes the BGP unnumbered test `AfterAll` cleanup to reset the ethernet interface instead of removing it. Previously, the cleanup used `WithAbsentInterface` which would disable/delete the interface entirely, potentially leaving the node in an unhealthy network state for subsequent tests. The fix switches to `WithEthernetDualStackInterface`, which restores the interface to its default dual-stack configuration.


evgenLevin (approved):
/lgtm

PR #1569, opened 2026-08-06
ocp sriov: fix flaky metrics exporter test by handling transient empt…

evgenLevin:
…y PromQL responses

Replace hard Expect assertions in fetchScalarFromPromQLoutput with graceful error handling that returns -1 on transient failures (empty output, invalid JSON, missing metrics). Since this function is only called within Eventually retry loops, returning -1 allows the test to retry rather than immediately failing on transient Prometheus gaps.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty, malformed, incomplete, and invalid metric responses.
  * Metric polling now retries when values cannot be parsed instead of failing immediately.
  * Added clearer logging for invalid monitoring data.


ajaggapa (approved):
/lgtm

PR #1575, opened 2026-08-07
global: add nmstate to eco-gotests container

klaskosk:
This commit updates the eco-gotests container to include the nmstate package. It is necessary for creating the IBI preinstall ISO and only increases container size by about 1% so the impact is generall insignificant.

Assisted-by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Added `nmstate` to the Docker image, enabling network state management capabilities.


yliu127 (approved):
Thanks!

PR #1579, opened 2026-08-12
cnf ran ptp: add comment explaining chronyc command

klaskosk:
This commit is documentation-only, adding a comment that explains the chronyc sources command that gets run to verify successful failover. The significance of the `^*` check is non-obvious, so it helps to be clear why this is what we assert.

Assisted-by: Cursor

## Summary by CodeRabbit

* **Documentation**
  * Added explanatory comments clarifying the NTP fallback check and the meaning of the selected-source indicator.


dpopsuev (approved):
LGTM, thanks for the doc!

PR #1583, opened 2026-08-14
cnf ran oran: remove test case 78246

klaskosk:
This commit removes O-RAN pre-provision test case 78246 and its associated helpers and parameters. The ability to provide inline BMC details was removed in openshift-kni/oran-o2ims#3233 so the code path this test case covered no longer exists. There is no similar code path where one can avoid the hardware manager anymore, so there is no possible way to update the test case to be useful.

Assisted-by: Cursor

## Summary by CodeRabbit

* **Tests**
  * Removed an outdated pre-provisioning test for cluster instances using inline BMC data.
  * Removed related test-only helpers and constants no longer needed by the test suite.

### DIFF
--- a/tests/cnf/ran/oran/internal/helper/helper.go
+++ b/tests/cnf/ran/oran/internal/helper/helper.go
@@ -3,7 +3,6 @@ package helper
 import (
 	"context"
 	"fmt"
-	"slices"
 	"strings"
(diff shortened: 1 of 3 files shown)


josclark42 (approved):
LGTM

PR #1597, opened 2026-08-24
ptp: Add dual T-BC time receiver ports (same NIC) holdover tests (CNF-23885)

hhassid:
## Summary

Adds automated tests for the dual time receiver ports (same NIC) T-BC feature (CNF-22709).
When a T-BC profile has two upstream time receiver ports on the same NIC (same PHC), ptp4l
uses priority-based failover between them. These tests validate holdover behavior for that
configuration on both WPC (E810) and GNRD (E825) environments.

Jira: https://redhat.atlassian.net/browse/CNF-23885

### Changes

- **New profile type `ProfileTypeDualTBCReceiver`** — classified by `clockType: T-BC` with
  2+ client interfaces in ptp4lConf, distinguished from two-port OC before the existing
  `ProfileTypeTwoPortOC` match.
- **`GetUpstreamPortsForProfile`** — new helper that splits comma-separated `upstreamPort`
  values (e.g. `ens7f1,ens7f3`) into multiple `iface.Name` entries. Single-port callers
  use the existing `GetUpstreamPortForProfile` which now delegates to it.
- **Holdover helpers refactored** — `ptp-holdover.go` helpers (`setUpstreamInterfacesStatus`,
  `restoreInterfacesAndWaitForRelock`) now operate on `[]iface.Name` so they work for both
  single and dual upstream port configurations.
- **New test file `ptp-dual-tbc.go`** with label `dual-tbc`, gated from PTP operator 5.0:
  - **OCP-89744** — Active TR port down → HOLDOVER (clock class 135) → failover to backup
    port → LOCKED (clock class 6). Validates interface roles (FAULTY / FOLLOWER).
  - **OCP-89746** — Both TR ports down → HOLDOVER → FREERUN (clock class 248). Both
    interfaces FAULTY. Restore → LOCKED.
  - Holdover in-spec → locked (no out-of-spec, IDs TBD)
  - Holdover in-spec → out-of-spec (IDs TBD)
  - Holdover out-of-spec → freerun (IDs TBD)
- **Unit tests** for `isDualTBCReceiver` profile classification and `GetUpstreamPortsForProfile`
  comma splitting.

### Notes

- OCP-89745 (BC, non T-BC counterpart) is intentionally not included per test plan.
- The existing single T-BC holdover tests (83297–83300) and T-TSC tests (88274–88277) are
  unaffected — they look up `ProfileTypeTBCReceiver` / `ProfileTypeTTSC` and skip when only
  a dual T-BC config is present, and vice versa.

## Test plan

- [x] Run on a **WPC** environment with a dual T-BC PtpConfig
      (e.g. `PtpConfigTbcTimeReciverPortsSameNic.yaml` — E810, `upstreamPort: ens7f1,ens7f3`)
      Label filter: `--ginkgo.label-filter="dual-tbc"`
- [x] Run on a **GNRD** environment with a dual T-BC PtpConfig
      (e.g. `PtpConfigDualTBcTRGnrd.yaml` — E825, `upstreamPort: eno8503np2,eno8603np3`)
      Label filter: `--ginkgo.label-filter="dual-tbc"`
- [x] Verify single T-BC holdover tests (label `tbc-tsc-holdover`) still pass on a
      cluster with a classic single-port T-BC config (no regression)
- [x] Verify dual T-BC tests skip cleanly on a cluster with no dual T-BC config

## Summary by CodeRabbit

* **New Features**
  * Added support for Dual T-BC receiver profiles with active and backup upstream interfaces.
  * Added failover and holdover validation scenarios for Dual T-BC configurations.
  * Added support for discovering and managing multiple upstream interfaces.
  * Added bulk interface state changes with detailed failure reporting.

* **Bug Fixes**
  * Improved passive-interface detection for listening and passive roles.
  * Updated holdover validation to support multiple upstream interfaces and clock-class variations.

* **Tests**
  * Added coverage for Dual T-BC profile detection, upstream-port parsing, failover, and holdover behavior.

### DIFF
--- a/tests/cnf/ran/ptp/internal/profiles/parser.go
+++ b/tests/cnf/ran/ptp/internal/profiles/parser.go
@@ -209,6 +209,10 @@ func determineProfileType(
 	// If the profile has one interface and one client interface, return ProfileTypeOC.
 	case numInterfaces == 1 && numClientInterfaces == 1:
 		return ProfileTypeOC, nil
+	// Dual T-BC must be classified before two-port OC: both have two client interfaces.
+	case numClientInterfaces == 2 && numServerInterfaces == 0 &&
+		isDualTBCReceiver(profile, controlledNames, profileName):
+		return ProfileTypeDualTBCReceiver, nil
 	// If the profile has two interfaces and two client interfaces, return ProfileTypeTwoPortOC.
 	case numInterfaces == 2 && numClientInterfaces == 2:
 		return ProfileTypeTwoPortOC, nil
@@ -222,6 +226,30 @@ func determineProfileType(
 	default:
 		return 0, fmt.Errorf("unable to determine PTP profile type based on defined rules")
 	}
+}
+
+// isDualTBCReceiver reports whether a multi-client profile is a dual time-receiver T-BC rather than a two-port OC.
+// Dual T-BC is identified by clockType T-BC, a controllingProfile reference, or a comma-separated upstreamPort
+// combined with telecom-slave (ts2phc/phc2sys) configuration.
+func isDualTBCReceiver(profile ptpv1.PtpProfile, controlledNames map[string]bool, profileName string) bool {
+	if profile.PtpSettings != nil && profile.PtpSettings["clockType"] == "T-BC" {
+		return true
+	}
+
+	if controlledNames[profileName] {
+		return true
+	}
+
+	return hasTelecomSlaveConfig(profile) && hasDualUpstreamPort(profile)
+}
+
+// hasDualUpstreamPort reports whether ptpSettings.upstreamPort lists more than one interface.
+func hasDualUpstreamPort(profile ptpv1.PtpProfile) bool {
+	if profile.PtpSettings == nil {
+		return false
+	}
+
+	return strings.Contains(profile.PtpSettings["upstreamPort"], ",")
 }
 
 // hasClientFlag checks if the ptp4lOpts string contains any client-only flags. Though the reference PTP profiles use
--- a/tests/cnf/ran/ptp/tests/ptp-dual-tbc.go
+++ b/tests/cnf/ran/ptp/tests/ptp-dual-tbc.go
@@ -1,0 +1,189 @@
+package tests
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/reportxml"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/internal/querier"
+	. "github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/internal/raninittools"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/internal/ranparam"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/internal/version"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/iface"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/metrics"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/profiles"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/ptp/internal/tsparams"
+	"k8s.io/klog/v2"
+)
+
+var _ = Describe("PTP Dual T-BC", Label(tsparams.LabelDualTBC), func() {
+	var prometheusAPI prometheusv1.API
+
+	BeforeEach(func() {
+		var err error
+
+		By("creating a Prometheus API client")
+
+		prometheusAPI, err = querier.CreatePrometheusAPIForCluster(RANConfig.Spoke1APIClient)
+		Expect(err).ToNot(HaveOccurred(), "Failed to create Prometheus API client")
+
+		By("checking if PTP operator version supports dual T-BC tests")
+
+		inRange, err := version.IsVersionStringInRange(
+			RANConfig.Spoke1OperatorVersions[ranparam.PTP], "5.0.0-0", "")
+		Expect(err).ToNot(HaveOccurred(), "Failed to parse PTP operator version")
+
+		if !inRange {
+			Skip("Test is valid from version 5.0")
+		}
+	})
+
+	Context("dual t-bc dual time receiver ports on the same NIC", func() {
+		var testData profiles.HoldoverTestData
+
+		timeout := holdoverTestTimeout
+
+		BeforeEach(func() {
+			By("getting node info map")
+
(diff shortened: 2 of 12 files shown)


tests/cnf/ran/ptp/internal/profiles/parser.go
```diff
@@ -209,6 +209,10 @@ func determineProfileType(
 	// If the profile has one interface and one client interface, return ProfileTypeOC.
 	case numInterfaces == 1 && numClientInterfaces == 1:
 		return ProfileTypeOC, nil
+	// Dual T-BC must be classified before two-port OC: both have two client interfaces.
+	case numClientInterfaces >= 2 && numServerInterfaces == 0 &&
+		isDualTBCReceiver(profile, controlledNames, profileName):
```
klaskosk, line 214: could we simplify by mirroring the T-BC receiver logic?

```go
numInterfaces == 2 && numClientInterfaces == 2 && controlledNames[profileName]
```

tests/cnf/ran/ptp/tests/ptp-holdover.go
```diff
@@ -383,41 +369,55 @@ func assertHoldoverOutOfSpecToFreerun(
...
-	Expect(err).ToNot(HaveOccurred(), "Failed to set upstream clock interface up")
+	setUpstreamInterfacesStatus(testData, iface.InterfaceStateUp)
 
 	assertLockedState(testData.prometheusAPI, testData.nodeName, ifaceUpTime,
 		expected.Locked, clockClassChanges, timeout)
 }
 
-// restoreInterfaceAndWaitForRelock brings the upstream interface back up and waits for the T-BC clock to return
+// setUpstreamInterfacesStatus sets every upstream interface on the test node to the given state.
+func setUpstreamInterfacesStatus(testData holdoverTestData, state iface.InterfaceState) {
```
klaskosk: it looks like this helper can be removed just by changing the way that `setInterfacesStatus` is called

tests/cnf/ran/ptp/tests/ptp-dual-tbc.go
```diff
@@ -0,0 +1,197 @@
...
+		inRange, err := version.IsVersionStringInRange(
+			RANConfig.Spoke1OperatorVersions[ranparam.PTP], "5.0.0-0", "")
+		Expect(err).ToNot(HaveOccurred(), "Failed to parse PTP operator version")
+
+		if !inRange {
+			Skip("Test is valid from version 5.0")
+		}
+	})
+
+	Context("dual t-bc dual time receiver ports on the same NIC", func() {
```
klaskosk, line 45: i wonder if it's worth moving these tests into the ptp-holdover file? they're written very similarly and rely on similar helpers, so it could make sense to have them in one place

also, would we want to run the T-BC tests from ptp-holdover for this config? since the PR updates the behavior to shut down all receiver interfaces, it seems like the behavior would be the same then

tests/cnf/ran/ptp/tests/ptp-dual-tbc.go
```diff
@@ -0,0 +1,197 @@
...
+	Expect(err).ToNot(HaveOccurred(),
+		"Failed to determine active/backup interfaces for dual T-BC on node %s", testData.nodeName)
+
+	By(fmt.Sprintf("identified active interface: %s, backup interface: %s", activeIface, backupIface))
+
+	return activeIface, backupIface
+}
+
+// assertInterfaceRole waits for the given interface to report the expected PTP role.
+func assertInterfaceRole(
```
klaskosk: i'm not personally a fan of the helpers that directly wrap the prometheus query assertions

although we've already opened pandora's box here in the holdover tests so it's not that big a deal, just a preference

klaskosk (commented):
a few pedantic comments 🙂, but I think the main consideration is whether dual port T-BC receiver should run the single port tests too. It feels like the config should function the same way on holdover->freerun when both ports are disconnected, but I could be wrong about this. it also could be a little redundant to test, so I'm curious what your thoughts are

tests/cnf/ran/ptp/tests/ptp-dual-tbc.go
```diff
@@ -0,0 +1,197 @@
...
+			reportxml.ID("89744"), func() {
+				assertDualTBCFailoverToBackup(testData, timeout)
+			})
+
+		// 89746 - Validates HOLDOVER and FREERUN after both ports are disconnected.
+		It("validates dual t-bc holdover and freerun after both ports are disconnected",
+			reportxml.ID("89746"), func() {
+				assertDualTBCHoldoverInSpecToFreerun(testData, holdoverPluginSettingsNoOutOfSpec,
+					timeout, TBCClockClasses())
+			})
```
DanyPops: How much of an effort would it take to collapse this test case scenario with the existing Locked -> Holdover -> Freerun?

tests/cnf/ran/ptp/tests/ptp-dual-tbc.go
hhassid: It'a a question I had in my mind and the first thought was to wright those in ptp-holdover.
2 reasons way not is we have oc-2-port which is similar to dual-t-bc, and the second that the ptp-holdover is already very long

tests/cnf/ran/ptp/internal/profiles/parser.go
hhassid: Agree

tests/cnf/ran/ptp/tests/ptp-holdover.go
hhassid: Agree, removed

tests/cnf/ran/ptp/tests/ptp-dual-tbc.go
dpopsuev: Ditto; looking at 89746 it appears to be a logical clone, @hhassid assess how much effort would it be reform the existing test cases so T-BC Dual resuses the same logic. This is an area of the code I'm tackling at the moment to simplify deduping.

tests/cnf/ran/ptp/tests/ptp-dual-tbc.go
hhassid: OK, there is no good point to do so in that case so I removed it and the interface-role assertion moved to `It`

tests/cnf/ran/ptp/tests/ptp-dual-tbc.go
hhassid: I understand there are duplications.
We can see same test flow in T-BC and T-TSC
Dual-T-BC and OC-2-ports also have similarly that can be merged.
We tried this approach in the beginning of ptp tests in cnf-gotests and I think it's better to have this approach although we have duplications. If we have a strong infrastructure and build same tests and only make a distinguish between different profiles.

klaskosk: fair, the oc-2-port example is why I wasn't strongly leaning either way. we can keep separate files then

+1 for the renaming of the OC2Port function, it makes more sense being a generic name now

@hhassid my other question then is whether we should run ptp-holdover for the dual BC receiver profiles?

hhassid: @klaskosk 
Agree, I think we should. But let's first focus on the polarion test cases. It was my mistake not to create those on time.

klaskosk (approved):

PR #1538, opened 2026-07-28
cnf network: fix bgpunnumbered afterAll

gkopels:
BGP Unnumbered intentionally “disabled” ECO_CNF_CORE_NET_SRIOV_INTERFACE_LIST[0] in teardown. Deleting the NNCP does not undo that apply, so the PF stayed down for subsequent tests in the same job/cluster. This cleanup has existed since the suite was added (#571); the failure shows up as cross-suite pollution, not a regression inside BGP Unnumbered itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup after BGP Unnumbered network tests by restoring the worker’s Ethernet interface to an active state.
  * Ensured network policies are removed even when interface restoration encounters an error.
  * Reported interface restoration and policy cleanup errors separately.

PR #1613, opened 2026-08-31
cnf network: fix unnumbered nmstate 4-21

gkopels:
- BGP Unnumbered AfterAll was resetting `ens8f0np0` with `WithEthernetDualStackInterface` and the `workercnf` label, so the NNCP applied to every worker. worker2 failed (`ipv4.enabled desire 'true', current 'false'`), the policy stayed Degraded, and AfterAll hit `context deadline exceeded`.
- Restore the PF on the setup node only (hostname selector) and leave it up with `WithEthernetIPv6LinkLocalInterface`. Do not use `WithAbsentInterface` (that leaves the PF down after NNCP deletion and breaks later suites) and do not force IPv4. 4.21 eco-goinfra has no `WithInterfaceUp` (main #1538).
- Capture the restore error so `CleanAllNMStatePolicies` still runs if restore fails.
- Same change as release-4.20 #1612.


ajaggapa (approved):
/lgtm

PR #1614, opened 2026-08-31
cnf network: fix unnumbered nmstate 4-22

gkopels:
- BGP Unnumbered AfterAll was resetting the SR-IOV PF with `WithEthernetDualStackInterface` and the `workercnf` label, so the NNCP applied to every worker. That can leave the policy Degraded (`ipv4.enabled desire 'true', current 'false'`) and AfterAll hits `context deadline exceeded`.
- Restore the PF on the setup node only (hostname selector) and leave it up with `WithInterfaceUp`. Do not use `WithAbsentInterface` (that leaves the PF down after NNCP deletion and breaks later suites).
- Capture the restore error so `CleanAllNMStatePolicies` still runs if restore fails.
- Same cleanup as main #1538 / release-4.20 #1612.


ajaggapa (approved):
/lgtm

PR #1625, opened 2026-09-03
cnf network: fix networkpolicy-flake

gkopels:
DeferCleanup(enableMultiNetworkPolicy(true)) at the start of 55990 so the feature is turned back on even if the spec fails, before the next specs run. The explicit re-enable at the end of the spec is unchanged.

Longer post-disable TCP waits: after disable, the two “traffic should work again” Eventually checks use tsparams.WaitTimeout (3m) instead of WaitTrafficTimeout (1m). Deny-all checks before disable still use 1m.


evgenLevin (approved):
lgtm

ajaggapa (approved):

PR #1626, opened 2026-09-03
Draft: CNF-17455 improve basic ptp sync test case

Bonnie-Block:
Submitting for feedback.
Tested in:
GNR-D T-BC and GM
WPC BC/OC

### DIFF
--- a/tests/cnf/internal/nicinfo/nicinfo.go
+++ b/tests/cnf/internal/nicinfo/nicinfo.go
@@ -166,16 +166,18 @@ const (
 	ethtoolRetryInterval = 20 * time.Second
 )
 
-// execEthtoolWithRetryOnEmpty executes an ethtool command on a node and retries if the output is empty.
-// This is necessary because ethtool output can sometimes be inexplicably empty.
-func execEthtoolWithRetryOnEmpty(
-	client *clients.Settings, nodeName, command string, nodeSelector metav1.ListOptions) (string, error) {
+// execEthtoolWithRetry executes an ethtool command on a node and retries when output is empty or
+// cannot be parsed. cluster.ExecCmdWithStdout already normalizes exec output via execoutput.Normalize.
+func execEthtoolWithRetry(
+	client *clients.Settings, nodeName, command string, nodeSelector metav1.ListOptions,
+	parse func(string) error,
+) (string, error) {
 	var output string
 
 	for attempt := range ethtoolRetries {
 		if attempt > 0 {
 			klog.V(logLevel).Infof(
-				"Retrying ethtool command %q on node %s (attempt %d/%d) due to empty output",
+				"Retrying ethtool command %q on node %s (attempt %d/%d)",
 				command, nodeName, attempt+1, ethtoolRetries)
 			time.Sleep(ethtoolRetryInterval)
 		}
@@ -187,15 +189,49 @@ func execEthtoolWithRetryOnEmpty(
 		}
 
 		output = outputs[nodeName]
-		if strings.TrimSpace(output) != "" {
-			return output, nil
+		if strings.TrimSpace(output) == "" {
+			klog.V(logLevel).Infof("Ethtool command %q on node %s returned empty output", command, nodeName)
+
+			continue
 		}
 
-		klog.V(logLevel).Infof("Ethtool command %q on node %s returned empty output", command, nodeName)
+		if err := parse(output); err != nil {
+			klog.V(logLevel).Infof(
+				"Ethtool command %q on node %s returned unparseable output (attempt %d/%d): %v",
+				command, nodeName, attempt+1, ethtoolRetries, err)
+
+			continue
+		}
+
+		return output, nil
 	}
 
-	return "", fmt.Errorf("ethtool command %q on node %s returned empty output after %d attempts",
-		command, nodeName, ethtoolRetries)
+	return "", fmt.Errorf("ethtool command %q on node %s failed after %d attempts: last output was %q",
+		command, nodeName, ethtoolRetries, output)
+}
+
+func parseDriverInfoOutput(output string) error {
+	if len(driverRegex.FindStringSubmatch(output)) == 0 {
+		return fmt.Errorf("driver line not found")
+	}
+
+	if len(versionRegex.FindStringSubmatch(output)) == 0 {
+		return fmt.Errorf("version line not found")
+	}
+
+	if len(firmwareVersionRegex.FindStringSubmatch(output)) == 0 {
+		return fmt.Errorf("firmware-version line not found")
+	}
+
+	return nil
+}
+
+func parsePTPHardwareClockOutput(output string) error {
+	if len(ptpHardwareClockRegex.FindStringSubmatch(output)) == 0 {
+		return fmt.Errorf("PTP hardware clock line not found")
+	}
+
+	return nil
 }
 
 // getInterfaceInfo gets the information for a given interface on a given node by running ethtool commands on the node
@@ -208,7 +244,8 @@ func getInterfaceInfo(client *clients.Settings, nodeName string, interfaceName s
 
 	driverInfoCommand := fmt.Sprintf("ethtool -i %s", interfaceName)
 
-	driverInfoOutput, err := execEthtoolWithRetryOnEmpty(client, nodeName, driverInfoCommand, nodeSelector)
+	driverInfoOutput, err := execEthtoolWithRetry(
+		client, nodeName, driverInfoCommand, nodeSelector, parseDriverInfoOutput)
 	if err != nil {
 		return nicInfo{}, fmt.Errorf(
 			"failed to get driver info for interface %s on node %s: %w", interfaceName, nodeName, err)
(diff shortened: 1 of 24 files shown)


tests/internal/execoutput/execoutput.go
```diff
@@ -0,0 +1,60 @@
+// Package execoutput normalizes stdout from remote exec (MCO nsenter, linuxptp-daemon pod)
```
klaskosk: i think this is a good change, although I am a little concerned rolling it out so broadly since so many different teams use the command execution functions

tests/cnf/ran/ptp/internal/processes/processes.go
```diff
@@ -221,7 +228,7 @@ func WaitForProcessRunning(
 // getRelatedIndices returns the indices of the related processes for a given process.
 // error is nil if the matches are found.
 func getRelatedIndices(client *clients.Settings, nodeName string, relatedProcess PtpProcess) ([]string, error) {
-	getIndexCommand := fmt.Sprintf("pgrep -a %s | grep /var/run |  head -n 1", relatedProcess)
+	getIndexCommand := fmt.Sprintf("ls -1 /var/run/%s.*.config 2>/dev/null | head -n 1", relatedProcess)
```
klaskosk: i'm not sure the reason for changing the command here. is there an edge case which causes these to be different that we're trying to capture?

tests/cnf/ran/ptp/internal/metrics/expected.go
```diff
@@ -0,0 +1,85 @@
...
+// deduplicateExpectedClockStates removes duplicate entries from the expected clock states slice. Two entries
+// are considered duplicates if they have the same process, interface, and node.
+func deduplicateExpectedClockStates(expected []ExpectedClockState) []ExpectedClockState {
+	type key struct {
+		process PtpProcess
+		iface   string
+		node    string
+	}
+
+	seen := make(map[key]struct{})
```
klaskosk, line 70: nit: this could be `map[ExpectedClockState]struct{}`

tests/cnf/ran/ptp/internal/metrics/ensurelocked.go
```diff
@@ -6,23 +6,108 @@ import (
...
+	"k8s.io/klog/v2"
 )
 
+const (
+	defaultLockedStableDuration = 10 * time.Second
+	defaultLockedTimeout        = 5 * time.Minute
+)
+
+// ensureLockedConfig holds the configuration for EnsureClocksAreLocked and EnsureClocksAreStable.
+type ensureLockedConfig struct {
```
klaskosk, line 19: in general, I think the requirement for more options in this file is a sign we shouldn't be using it rather than a sign to add more options

it's my fault for starting it since it was always a hack to workaround behavior that's now considered a bug, but I think it's at its limit

since by adding these options and configuration, we're really just creating a new wrapper around assertions that has only some of the functionality

tests/internal/cluster/cluster.go
```diff
@@ -184,7 +185,7 @@ func ExecCmdWithStdout(
 			}
 
 			hostname := regexp.MustCompile(`\r`).ReplaceAllString(hostnameBuf.String(), "")
-			output := regexp.MustCompile(`\r`).ReplaceAllString(commandBuf.String(), "")
+			output := execoutput.Normalize(commandBuf.String())
```
klaskosk, line 188: from meeting: let's remove the Normalize from here (leave in the ptpdaemon package) and move execoutput to its own PR

tests/cnf/internal/nicinfo/nicinfo.go
```diff
@@ -166,16 +166,18 @@ const (
 	ethtoolRetryInterval = 20 * time.Second
 )
 
-// execEthtoolWithRetryOnEmpty executes an ethtool command on a node and retries if the output is empty.
-// This is necessary because ethtool output can sometimes be inexplicably empty.
-func execEthtoolWithRetryOnEmpty(
-	client *clients.Settings, nodeName, command string, nodeSelector metav1.ListOptions) (string, error) {
+// execEthtoolWithRetry executes an ethtool command on a node and retries when output is empty or
```
klaskosk, line 169: from meeting: good candidate to be part of a separate PR, kirsten was not aware there was another automation issue here but changes look good

tests/cnf/ran/ptp/internal/processes/processes.go
```diff
@@ -115,30 +116,36 @@ func GetPtp4lPIDsByRelatedProcess(
 		return nil, err
 	}
 
-	grepFlag := ""
-	if !related {
-		grepFlag = "-v"
-	}
+	indicesPattern := strings.Join(indices, "|")
```
klaskosk, line 119: from meeting: there's cases where a process doesn't restart properly and we don't notice due to how we check (or in the past, a process never started to begin with)

please add explanation comment for why we're doing it this way (catching errors the old method didn't)

tests/cnf/ran/ptp/internal/metrics/ensurelocked.go
```diff
@@ -6,23 +6,108 @@ import (
...
-// which is tested separately, not the steady state this function validates.
-func EnsureClocksAreLocked(prometheusAPI prometheusv1.API) error {
-	query := ClockStateQuery{
-		Process: DoesNotEqual(ProcessChronyd),
+// By default it uses the broad cluster-wide locked baseline ([LockedClockExpectations]). When
+// [WithExpectedClockStates] is provided, it asserts each expected series is present and LOCKED.
+//
+// It ensures that clocks are locked for 10 seconds with a timeout of 5 minutes unless overridden by
+// [WithAssertOptions].
+func EnsureClocksAreLocked(prometheusAPI prometheusv1.API, opts ...EnsureLockedOption) error {
```
klaskosk, line 83: from meeting: let's keep the EnsureClocksAreLocked pretty much the same since it's used more broadly as a basic check to ensure the cluster is in decent shape in between tests

for events-and-metrics (sync tests), we can do the more complex checks of ensuring all the clock states we expect to see exist and are locked

tests/cnf/ran/ptp/internal/profiles/expectedmetrics.go
```diff
@@ -0,0 +1,271 @@
...
+		return true
+	case ProfileTypeOC, ProfileTypeTwoPortOC, ProfileTypeBC, ProfileTypeHA,
+		ProfileTypeTBCTransmitter, ProfileTypeTBCReceiver, ProfileTypeDualTBCReceiver, ProfileTypeTTSC:
+		return false
+	}
+
+	return false
+}
+
+// ptp4lUsesNICName returns true if the profile type reports ptp4l clock state metrics with NIC names rather
```
klaskosk, line 148: from meeting: Daniel has a backlog story to use the operator's interface alias map endpoint (since it's more complex than just configs) so we can just ignore the issue until then

tests/cnf/ran/ptp/internal/metrics/assert_set.go
```diff
@@ -0,0 +1,136 @@
+package metrics
```
klaskosk, line 1: i dug up [my sketch of multi query assertions](https://github.com/klaskosk/eco-gotests/commit/2584ef9f1b3f886ad3e4e9e04619752f4bf65fb9) from the previous sync test changes. i made it concurrent, though that's not really necessary

of the main differences:

- i have a slight preference for storing the query/expected rather than the assertion function, though it's pretty similar in practice
- writing the single query version as a special case of the multi query version does allow sharing the options initialization and simplifies the code a bit
- i like your `Set` function, it fits in well with the assertions DSL we have

tests/cnf/ran/ptp/internal/metrics/assert_set_test.go
```diff
@@ -0,0 +1,419 @@
...
+	"testing"
+	"time"
+
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type fakePrometheusAPI struct {
```
klaskosk, line 19: i wish prometheus provided a fake package similar to k8s 🙁 

i don't love mocking the client, but unit testing the finer points of our assertions is good

tests/internal/execoutput/execoutput_test.go
```diff
@@ -0,0 +1,55 @@
...
+	raw := "\x1b[1;31m2026-09-02T22:29:13.122647Z: driver: ice\n" +
+		"version: 5.14.0-687.42.1.el9_8.x86_64+rt\n" +
+		"firmware-version: 1.22 0x8001873d 1.3909.0\n" +
+		"supports-register-dump: yes\nsup\x1b[0m\n" +
+		"\x1b[1;31m2026-09-02T22:29:13.122701Z: ports-priv-flags: yes\n\x1b[0m\n"
+
+	normalized := Normalize(raw)
+	if !stringsHasPrefixLine(normalized, "driver: ice") {
+		t.Fatalf("expected driver line after normalize, got %q", normalized)
+	}
```
klaskosk, line 20: ah I missed this yesterday, but we should use the `testify` package to do assertions. AI seems pretty bad at figuring that out without saying it explicitly 🙃

klaskosk (commented):
i took a closer look at the metrics changes and added a comment there, plus a minor one about testify

PR #1631, opened 2026-09-05
Enhance TALM test cases with validation of expected events

dgonyier:
Check for TALM event types and generation with  events v1 API calls

Currenlty failing tests (3) have bugs posted. Other tests pass.

## Summary by CodeRabbit

* **Tests**
  * Expanded ClusterGroupUpgrade event validation across batching, blocking, and canary scenarios.
  * Added coverage for lifecycle, timeout, validation-failure, scope, and annotation details.
  * Improved event polling, ordering checks, cleanup, and diagnostic reporting.
  * Added typed test parameters for event reasons, scopes, and annotations.
  * Included event resources in collected test reports for troubleshooting.


tests/cnf/ran/talm/internal/tsparams/consts.go
```diff
@@ -139,6 +139,35 @@ const (
...
+	CguRegardingKind = "ClusterGroupUpgrade"
+	// CguMissingClustersAnnotation is annotation key for missing clusters list.
+	CguMissingClustersAnnotation = "cgu.openshift.io/missing-clusters"
+	// CguMissingClustersCountAnnotation is annotation key for missing cluster count.
+	CguMissingClustersCountAnnotation = "cgu.openshift.io/missing-clusters-count"
+	// CguMissingPoliciesAnnotation is annotation key for missing policies list.
+	CguMissingPoliciesAnnotation = "cgu.openshift.io/missing-policies"
+	// CguTimedoutClustersAnnotation is annotation key for timed-out clusters list.
+	CguTimedoutClustersAnnotation = "cgu.openshift.io/timedout-clusters"
+	// CguCreated is event reason when CGU is created.
```
klaskosk, line 164: it would've been nice if these constants were exported from the operator's api

the operator doesn't do this internally, but it could be good to separate the constants into distinct types to make them function more as enums

tests/cnf/ran/talm/tests/talm-batching.go
```diff
@@ -72,6 +81,43 @@ var _ = Describe("TALM Batching Tests", Label(tsparams.LabelBatchingTestCases),
...
+
+				if !helper.HasEventWithAnnotation(events, tsparams.CguMissingClustersCountAnnotation) {
+					return false, fmt.Errorf("missing missing-clusters-count annotation on validation event")
+				}
+
+				return true, nil
+			}).WithTimeout(30*time.Second).WithPolling(5*time.Second).Should(BeTrue(),
+				"[EVENT CHECK] Missing CguValidationFailure event or missing-clusters annotations")
+
+			// Skip: OCPBUGS-120826 — TALM does not set scope annotation on validation events.
```
klaskosk: it's fine for draft stage, but we should include these checks when we merge, even if they fail

tests/cnf/ran/talm/tests/talm-batching.go
```diff
@@ -200,6 +309,27 @@ var _ = Describe("TALM Batching Tests", Label(tsparams.LabelBatchingTestCases),
...
 			Expect(err).ToNot(HaveOccurred(), "Failed to wait for CGU to timeout")
 
+			By("verifying CGU emitted timeout events")
+
+			expectedSequence := []helper.EventMatcher{
+				{Reason: tsparams.CguTimedout, Scope: tsparams.EventScopeBatch},
+				{Reason: tsparams.CguTimedout, Scope: tsparams.EventScopeGlobal},
+			}
+
+			Eventually(func() (bool, error) {
```
klaskosk: since it shows up often, a helper for the asynchronous assertion of event sequences would be good

tests/cnf/ran/talm/tests/talm-batching.go
```diff
@@ -72,6 +81,43 @@ var _ = Describe("TALM Batching Tests", Label(tsparams.LabelBatchingTestCases),
...
+					return false, fmt.Errorf("no CguValidationFailure events found")
+				}
+
+				if !helper.HasEventWithAnnotation(events, tsparams.CguMissingClustersAnnotation) {
+					return false, fmt.Errorf("missing missing-clusters annotation on validation event")
+				}
+
+				if !helper.HasEventWithAnnotation(events, tsparams.CguMissingClustersCountAnnotation) {
+					return false, fmt.Errorf("missing missing-clusters-count annotation on validation event")
+				}
```
klaskosk: should we be checking on the `validationEvents` array instead of `events`? the error message suggests so, otherwise we get a slightly weaker assertion that the annotation appears, not that it appears on a validation event

tests/cnf/ran/talm/tests/talm-batching.go
```diff
@@ -92,6 +138,39 @@ var _ = Describe("TALM Batching Tests", Label(tsparams.LabelBatchingTestCases),
 			// This should immediately error out so we don't need a long timeout
 			_, err = cguBuilder.WaitForCondition(tsparams.CguNonExistentPolicyCondition, 2*time.Minute)
 			Expect(err).ToNot(HaveOccurred(), "Failed to wait for CGU to have matching condition")
+
+			By("verifying CGU emitted validation failure event with annotations")
+
+			Eventually(func() (bool, error) {
```
klaskosk: i think there's an opportunity to unify these assertions with the sequential ones, even though they're slightly different

the high level process is the same: polling until the list of events for a CGU satisfies some condition

and I think it's possible to generalize the event matching between these types of assertions and the sequential ones, since the matcher is just a way to determine whether a given event satisfies certain criteria

---

i think to implement this, you could change an event matcher to a function and provide constructors:

```go
type EventMatcher func(*eventsv1.Event) bool

func MatchReasonAndAnnotation(reason, annotation string) EventMatcher {
	return func(event *eventsv1.Event) bool {
		return event.Reason == reason && event.Annotations[annotation] != ""
	}
}
```

that'd also allow you to use `slices.IndexFunc` to determine if a matching event exists in a list

---

there's a few different routes you could go with this kind of approach, i think there's benefits even if you don't go as far as unifying the sequential assertions with the existence ones, since it'd allow you to generalize the polling into a helper

klaskosk (commented):
added a few comments, though they're mostly about the opportunity to use a helper to reduce some of the Eventually repetition

tests/cnf/ran/talm/tests/talm-batching.go
dgonyier: The code review skill report was to avoid submitting tests that will already fail and skip the failing parts until the bugs were fixed. I'm ok with keeping them failing.

tests/cnf/ran/talm/tests/talm-batching.go
```diff
@@ -131,6 +210,36 @@ var _ = Describe("TALM Batching Tests", Label(tsparams.LabelBatchingTestCases),
...
+				}
+
+				matched, seqErr := helper.VerifyEventSequence(events, expectedSequence)
+				if !matched {
+					return false, seqErr
+				}
+
+				if !helper.HasEventWithAnnotation(events, tsparams.CguTimedoutClustersAnnotation) {
+					return false, fmt.Errorf("missing timedout-clusters annotation on timeout events")
+				}
```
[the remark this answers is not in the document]
dgonyier: cursor:

This comment is not valid for TALM events. Here's why:

From the actual test event log data, TALM events:

Never have a series field — grep -c 'series' returns 0 across all 539 lines of event data
Each event is a distinct object with a unique generated name (e.g., talm-cgu-49k8n, talm-cgu-bwh66)
Each event type uses a different action value — RemediationStarted, RemediationInBatchStarted, RemediationTimeout, etc.
The Kubernetes Event.Series field is populated by the client-go event broadcaster when it deduplicates "isomorphic" events — events with identical Regarding, Action, Reason, ReportingController, ReportingInstance, and Related fields. TALM creates each event as a new API object with a unique generated name and distinct action values per event type, so deduplication never occurs.

The only test that uses Count: 2 is:


talm-batching.go
Ln 580–589
expectedSequence := []helper.EventMatcher{
	{Reason: tsparams.CguCreated, Scope: tsparams.EventScopeGlobal},
	{Reason: tsparams.CguStarted, Scope: tsparams.EventScopeGlobal},
	{Reason: tsparams.CguStarted, Scope: tsparams.EventScopeBatch},
	{Reason: tsparams.CguStarted, Scope: tsparams.EventScopeCluster, Count: 2},
	{Reason: tsparams.CguSuccess, Scope: tsparams.EventScopeCluster, Count: 2},
	{Reason: tsparams.CguSuccess, Scope: tsparams.EventScopeBatch},
	{Reason: tsparams.CguSuccess, Scope: tsparams.EventScopeGlobal},
}
Those two cluster-scoped events (for spoke1 and spoke2) have different action values (RemediationInClusterStarted vs potentially different actions per cluster), so they would remain as separate Event objects even if the broadcaster were involved.

Adding Series.Count handling would be dead code that never executes and adds complexity for no benefit. No changes needed.

tests/cnf/ran/talm/internal/tsparams/consts.go
dgonyier: should be addressed now

tests/cnf/ran/talm/tests/talm-batching.go
dgonyier: checks should now fail tests

tests/cnf/ran/talm/tests/talm-batching.go
dgonyier: helper added

tests/cnf/ran/talm/tests/talm-batching.go
dgonyier: should be addressed

tests/cnf/ran/talm/tests/talm-batching.go
dgonyier: should be addressed now

tests/cnf/ran/talm/internal/helper/cguevents.go
```diff
@@ -0,0 +1,307 @@
...
+				return value, true
+			}
+		}
+	}
+
+	return "", false
+}
+
+// CountEventsByReasonAndScope counts events matching both reason and scope.
+func CountEventsByReasonAndScope(
```
klaskosk: this function is unused

tests/cnf/ran/talm/internal/helper/cguevents.go
```diff
@@ -0,0 +1,307 @@
...
+		deleted++
+	}
+
+	klog.V(tsparams.LogLevel).Infof("Cleared %d CGU events in the %s namespace", deleted, tsparams.TestNamespace)
+
+	return nil
+}
+
+// FindEventsByReason filters events by reason, returning all matching events.
+func FindEventsByReason(events []*eventsv1.Event, reason tsparams.CguEventReason) []*eventsv1.Event {
```
klaskosk: this function is unused

tests/cnf/ran/talm/internal/helper/cguevents.go
```diff
@@ -0,0 +1,307 @@
...
+		if _, exists := event.Annotations[annotationKey]; exists {
+			return true
+		}
+	}
+
+	return false
+}
+
+// GetEventAnnotation retrieves annotation value from first event matching reason, returns value and exists flag.
+func GetEventAnnotation(
```
klaskosk: this function is unused

tests/cnf/ran/talm/tests/talm-batching.go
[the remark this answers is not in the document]
klaskosk: this analysis seems unrelated to what CodeRabbit is saying here

it's similar to my previous comment about validation events. the error says there exists an event in events such that the reason is timed out && the event has `CguTimedoutClustersAnnotation` but the actual assertion is that there exists an event in events such that the event has `CguTimedoutClustersAnnotation`

the latter is a weaker assertion than the error suggests

tests/cnf/ran/talm/internal/helper/cguevents.go
```diff
@@ -0,0 +1,272 @@
...
+	. "github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/internal/raninittools"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/cnf/ran/talm/internal/tsparams"
+)
+
+// EventMatcher defines expected event with reason, scope, and optional count.
+type EventMatcher struct {
+	Reason tsparams.CguEventReason // Event reason (e.g., CguStarted, CguSuccess)
+	Scope  tsparams.CguEventScope  // Event scope annotation (global, batch, cluster)
+	Count  int                     // Expected count: 0 = at least one, >0 = exact minimum
+	Strict bool                    // When true with Count>1, all occurrences must precede the next matcher
```
klaskosk: I ran a final AI pass of review and it looks like Strict isn't used and there are some implications to tests where Count > 1

klaskosk (commented):
1 more comment, otherwise lgtm

tests/cnf/ran/talm/internal/helper/cguevents.go
dgonyier: ~~remove Strict~~

Refactored Strict field handling instead

klaskosk (approved):
lgtm, thanks!

PR #1635, opened 2026-09-08
ptp: node reboot recovery scoped to the ptp, monitoring, ingress & API

dpopsuev:
Resolves ticket: https://redhat.atlassian.net/browse/CNF-26169

- cluster.WaitForRouteAPIAvailable(): Serves as a readiness guard for other calls to the 'Route' API.
- cluster.WaitForUnreachable(): Serves as a post-shutdown guard; stops a race in which the API is succcesfully probed right after shutdown.
- cosnt.OpenshiftAPINamespace: The OC API Namespace holds the API for 'Route' resources.

## Summary by CodeRabbit

- **Tests**
  - Improved PTP node reboot test coverage by validating cluster reachability and recovery after reboot.
  - Added checks that critical namespaces and pods return to a healthy state across supported cluster configurations.
  - Added route availability validation before stability tests interact with the Prometheus API.
  - Added validation for detecting temporarily unavailable clusters and confirming API route readiness.
  - Improved reliability of recovery checks following node disruptions.

### DIFF
--- a/tests/cnf/ran/ptp/tests/ptp-node-reboot.go
+++ b/tests/cnf/ran/ptp/tests/ptp-node-reboot.go
@@ -27,6 +27,13 @@ import (
 // This test must be ordered so that we only need to reboot the node once. It may continue on failure since the test
 // cases are not necessarily dependent on each other.
 var _ = Describe("PTP Node Reboot", Ordered, ContinueOnFailure, Label(tsparams.LabelNodeReboot), func() {
+	var testCriticalNs = []string{
+		ranparam.OpenshiftAPINamespace,
+		ranparam.OpenshiftIngressNamespace,
+		ranparam.PtpOperatorNamespace,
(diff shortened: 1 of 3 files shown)


tests/cnf/ran/ptp/tests/ptp-node-reboot.go
```diff
@@ -27,6 +27,13 @@ import (
 // This test must be ordered so that we only need to reboot the node once. It may continue on failure since the test
 // cases are not necessarily dependent on each other.
 var _ = Describe("PTP Node Reboot", Ordered, ContinueOnFailure, Label(tsparams.LabelNodeReboot), func() {
+	var testCriticalNs = []string{
```
klaskosk, line 30: this is a great idea, I know the original plan was just PTP namespaces like in cnf-gotests but we really do depend on others now

klaskosk (approved):
assuming it's been tested we can merge, lgtm!

PR #1636, opened 2026-09-08
lca seedgeneration: add authfile to podman and skopeo commands

klaskosk:
This commit adds the authfile argument to the podman and skopeo commands used by the seedgeneration suite. It reuses the mounted secret in the MCO pod, which comes from the cluster pull secret.

The existing tests assumed that the disconnected registry was readable without auth, but this is not always the case, necessitating this change.

Assisted-by: Cursor

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved seed image inspection and pulling by using the node’s registry authentication credentials.
  * Seed image verification now works more reliably with authenticated container registries.


yliu127 (approved):
lgtm

PR #1637, opened 2026-09-08
lca: redesign CNF TALM IBU suite around spoke health and fail-fast waits

yliu127:
Tests claim spoke health in each It, recover leftover IBU state only after a failed spec, and leave a successful upgrade intact for the ordered rollback test.

Add proper cluster health checks similarly as the LCA health check to the end of the test body - test pass means the cluster is completely ready for the next test case.

Upon test failure, various recovery actions will be performed depending on the state of the cluster. Examples including creating abort IBU and/or reboot node to recover sriov reset failure, etc. 

Expand log collection to all relevant components. e.g., upon test failure due to sriov node state unhealthy, the test will collect sriov daemon log which reveals the cause of the sriov node state issue.

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive spoke-cluster health checks covering nodes, operators, storage, backups, certificates, and upgrade resources.
  - Added automated recovery and readiness handling for interrupted, failed, aborted, or rolled-back image-based upgrades.
  - Enhanced upgrade reporting with additional hub and spoke resources, namespaces, and diagnostic information.
  - Added a pre-suite health check to verify spoke readiness before upgrade tests begin.

- **Bug Fixes**
  - Post-upgrade validations now run only when the upgrade succeeds.
  - Improved cleanup so recovery is attempted even when resource deletion encounters errors.

- **Refactor**
  - Standardized upgrade test setup, state tracking, waiting, cleanup, and recovery through shared helpers.

### DIFF
--- a/tests/lca/imagebasedupgrade/cnf/internal/cnfhelper/health.go
+++ b/tests/lca/imagebasedupgrade/cnf/internal/cnfhelper/health.go
@@ -1,0 +1,533 @@
+package cnfhelper
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/certificate"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/clients"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/clusteroperator"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/infrastructure"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/mco"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/nodes"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/oadp"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/olm"
+	oadpv1alpha1 "github.com/rh-ecosystem-edge/eco-goinfra/pkg/schemes/oadp/api/v1alpha1"
+	olmv1alpha1 "github.com/rh-ecosystem-edge/eco-goinfra/pkg/schemes/olm/operators/v1alpha1"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/sriov"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/velero"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/lca/imagebasedupgrade/cnf/internal/cnfinittools"
+	"github.com/rh-ecosystem-edge/eco-gotests/tests/lca/imagebasedupgrade/cnf/internal/cnfparams"
+	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	certificatesv1 "k8s.io/api/certificates/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// DefaultHealthTimeout is the overall timeout used by tests when claiming
+	// the spoke is healthy, including DefaultHealthStablePeriod.
+	DefaultHealthTimeout = 15 * time.Minute
+	// DefaultHealthStablePeriod is how long Check must stay healthy before
+	// WaitUntilHealthy succeeds. An unhealthy snapshot restarts this window.
+	DefaultHealthStablePeriod = 3 * time.Minute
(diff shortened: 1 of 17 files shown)


tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/upgrade-e2e-test.go
```diff
@@ -24,110 +22,115 @@ var _ = Describe(
...
+
+			By("Waiting until spoke IBU Rollback completes")
+
+			Expect(cnfhelper.WaitForSpokeIBU(cnfhelper.SpokeRollbackCompleted, 30*time.Minute)).To(Succeed(),
+				"Rollback did not complete on the spoke IBU")
+
+			By("Waiting until spoke IBU returns to Idle")
+
+			Expect(cnfhelper.WaitForSpokeIBU(cnfhelper.SpokeIdle, 10*time.Minute)).To(Succeed(),
+				"spoke IBU did not return to Idle after FinalizeRollback")
```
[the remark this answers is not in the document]
yliu127: The concern is that [Rollback, FinalizeRollback] plus a 5s poll can skip SpokeRollbackCompleted because LCA ResetStatusConditions wipes those conditions when Idle is reached.

LCA does not do that immediately:

TALM runs the two IBGU actions in order. FinalizeRollback cannot start until Rollback has completed on the spoke, so RollbackCompleted=True must exist for TALM to proceed.
When Idle is requested, LCA only sets Idle=False/Finalizing. RollbackCompleted stays True.
handleFinalize then runs cluster health checks (usually minutes after a rollback reboot), then ostree/OADP cleanup, and only then ResetStatusConditions.
So the wait is polling a condition that stays true for the whole finalize window, not a sub-5s blip. The test is also polling during the long rollback itself, so it should see RollbackCompleted well before finalize starts.

Dropping that wait and only accepting Idle would be worse: WaitForSpokeIBU(SpokeIdle) fail-fasts only on the Idle condition, not on RollbackCompleted=Failed. The two-step wait is what gives fail-fast on a failed rollback.

I would not change this. A reply on the thread explaining the LCA finalize sequence is enough.

tests/lca/imagebasedupgrade/cnf/internal/cnfhelper/health.go
```diff
@@ -0,0 +1,462 @@
...
+		return failedResult(checkNode, err.Error())
+	}
+
+	for _, nodeBuilder := range nodeList {
+		if result := snoNodeCheck(nodeBuilder); !result.Healthy {
+			return result
+		}
+	}
+
+	return passedResult(checkNode, "node is ready")
```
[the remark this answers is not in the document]
yliu127: not technically possible in IBU scenarios. Even if this returned empty list, the other health checks would have failed as well.

tests/lca/imagebasedupgrade/cnf/internal/cnfhelper/ibu.go
```diff
@@ -0,0 +1,1053 @@
...
+// if a stage reports Failed and that failure is not the expected end state.
+func WaitForSpokeIBU(expected SpokeIBUState, timeout time.Duration) error {
+	var lastStatus string
+
+	pollErr := wait.PollUntilContextTimeout(
+		context.TODO(), spokePollInterval, timeout, true,
+		func(_ context.Context) (bool, error) {
+			imageUpgrade, err := pullSpokeIBU()
+			if err != nil {
+				klog.Infof("waiting for spoke IBU %s: pull failed: %v", expected, err)
```
klaskosk: minor: logs should use `klog.V(cnfparams.CNFLogLevel)` instead of `klog.Infof` directly

goes for all uses of `klog`

tests/lca/imagebasedupgrade/cnf/upgrade-talm/tests/cleanup.go
```diff
@@ -0,0 +1,40 @@
...
+
+// leaveSuccessfulUpgrade reports whether Recover must not roll back a completed
+// Upgrade. That is only true after 68954 passed and before Ordered test 69058
+// runs; a failed 69058 or any independent spec must still recover to Idle.
+func leaveSuccessfulUpgrade() bool {
+	if !cnfibuvalidations.UpgradeSucceeded {
+		return false
+	}
+
+	return CurrentSpecReport().LeafNodeText != "Rollback successful upgrade"
```
klaskosk: minor: since we're matching on the string text, it'd be good to extract to a constant. that way they can't ever get out of sync and cause automation issues

tests/lca/imagebasedupgrade/cnf/internal/cnfhelper/ibu.go
```diff
@@ -0,0 +1,1053 @@
+package cnfhelper
```
klaskosk, line 1: optionally I think this file could be split up a little along wait/state/recover lines

the Wait functions feel pretty similar to what would be in eco-goinfra, but they're different enough that I think it makes sense to have them here

overall though the file seems pretty justified in its complexity, it's a complicated process to handle recovery and cleanup properly

yliu127: @klaskosk this has been split into ibu.go, ibgu.go, and recover.go. Ready for another review.

klaskosk (dismissed):

klaskosk (approved):

